PR #42, opened 2021-01-08
tooltips and zoom issue fixed

annika608:
added tooltips to legend, hover details on map, added zoom on segregation map, removed "click SA to view details about it" modal that was showing after clicking on an area

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,6 +24,7 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "^1.0.10",
     "react-sticky-footer": "0.1.0-rc3",
+    "react-tooltip":"0.0.1",
     "redux": "^4.0.5",
     "semantic-ui-react": "^0.88.2",
     "yarn": "^1.22.4"
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "^1.0.10",
     "react-sticky-footer": "0.1.0-rc3",
-    "react-tooltip":"0.0.1",
+    "react-tooltip": "^4.2.13",
     "redux": "^4.0.5",
     "semantic-ui-react": "^0.88.2",
     "yarn": "^1.22.4"
--- a/src/components/display.js
+++ b/src/components/display.js
@@ -2,13 +2,11 @@ import React from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
 import ReactTooltip from 'react-tooltip';
-import { Link} from 'react-router-dom';
 
 
 import * as Constants from "../constants";
 import { setFlowDirection, setDisplayDefault } from "../redux/action-creators";
 import BarGraph from "./BarGraph";
-import { Divider } from "semantic-ui-react";
 
 let Display = class Display extends React.Component {
   static propTypes = {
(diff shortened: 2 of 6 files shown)


src/components/display.js
```diff
@@ -1,10 +1,14 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { connect } from "react-redux";
+import ReactTooltip from 'react-tooltip';
```
justinanderson, line 4: I think you forgot to push your changes to package.json. react-tooltip is a new dependency.

src/components/display.js
```diff
@@ -1,10 +1,14 @@
...
+import { Link} from 'react-router-dom';
```
justinanderson, line 5: Link and Divider aren't used and should be removed here.

src/components/display.js
```diff
@@ -74,19 +78,44 @@ let Display = class Display extends React.Component {
         </div>
         <div className="py12 px12">
           <div className="mb6">
-            <h2 className="txt-bold txt-m block">GDP Growth Potential</h2>
-            <p className="txt-s">{ggp}</p>
+            <h2 className="txt-bold txt-m block" data-tip data-for = "GDPTip">GDP Growth Potential
+            {/* <Link  to = '/Research' style = {{ whiteSpace: "pre" }}>{`${'                      '}`} ? </Link>  */}
```
justinanderson, line 82: Rather than committing commented out code, it's better to leave those lines out of the commit or delete them entirely. Commented out code has a way of piling up, and people are usually timid about deleting others' commented lines, so it can make things cluttered over time.

justinanderson (changes requested):

package.json
```diff
@@ -24,6 +24,7 @@
     "react-router-dom": "^5.2.0",
     "react-scripts": "^1.0.10",
     "react-sticky-footer": "0.1.0-rc3",
+    "react-tooltip":"0.0.1",
```
justinanderson, line 27: Why react-tooltip version 0.0.1 instead of the latest version? Did you use `npm install react-tooltip --save` to add it to the project?

Using version 0.0.1 leads to this compilation error, which isn't present using the latest version:

```
./node_modules/react-tooltip/index.js
Module parse failed: Unexpected token (75:6)
You may need an appropriate loader to handle this file type.
|     })
|     return (
|       <span className={toolTipClass} style={style} data-id="tooltip">{this.state.placeholder}</span>
|     )
|   }
```

justinanderson (changes requested):

src/components/segregation.js
```diff
@@ -133,6 +145,20 @@ let SegregationMap = class SegregationMap extends React.Component {
...
+          style: "currency",
+          currency: "AUS",
+        })
+        this.hoveredPopup
+          .setLngLat(coordinates)
+          .setHTML("<h5>" + regionName +
+          "</h5> <p> <b> Population: </b> " + e.features[0].properties.persons_num + 
+          "<br /> <b> Median Income (AUS): </b>" + medIncome+ 
+          "<br / > <b> GDP Growth Potential: </b>" + e.features[0].properties.income_diversity+
+          "<br / > <b> Job Resiliance: </b>" + e.features[0].properties.bridge_diversity +"</p>" )          .addTo(this.map);
```
justinanderson, line 160: Compiler warning here. Looks like it's missing a newline before .addTo.
```
Line 154:  Unexpected whitespace before property addTo  no-whitespace-before-property
```

PR #54, opened 2021-01-22
Dropdown on map legend

LeoSilvaGomes:
- [x] Dropdown menu to map legend
- [x] Create gradient on legend


src/components/legend.js
```diff
@@ -10,27 +13,95 @@ let Legend = class Legend extends React.Component {
...
+    };
+
     let legendName = this.props.active.legendName;
     return (
-      <div className="bg-white absolute bottom left ml24 mb36 py12 px12 shadow-darken10 round z1 wmax180">
+      <div className="bg-white absolute bottom left ml24 mb36 py12 px12 shadow-darken10 round z1 wmax220">
+        <div className="mt6 mb12">
+          <Dropdown
+            defaultValue={mapType[0].value}
+            search
```
justinanderson, line 94: I think we don't want this to be searchable.

justinanderson (changes requested):

PR #78, opened 2021-04-05
Clean dependencies

justinanderson:
- Removes unused dependencies.
- Upgrades to the latest version of react-scripts.
- Adds the script `npm run build` for compiling production builds.


maitanavarro (commented):
I received this error when I tried to run it (not sure if this is a problem on my end)
Screen Shot 2021-04-06 at 3 25 14 PM

justinanderson:
@maitanavarro Try doing an `npm clean-install` first?

The mathjax dependency hasn't changed, so it's weird that you'd get an error there.

maitanavarro:
Hmm yeah, I used "npm ci" the first time before I did "npm start", and tried it again with "npm clean-install", and I still got the same error

justinanderson:
What version of node and npm do you have? On my computer, it's:

```sh
$ node --version
v12.13.1
$ npm --version
7.7.6
```

maitanavarro:
Oh! Just checked and my computer has:

```
$ node --version
v14.15.4
$ npm --version
6.14.10

justinanderson:
Ok, yeah. I can get that same error on my computer if I use node 14. It runs successfully if I delete package-lock.json and let `npm install` resolve dependencies from scratch.

I forget which ones, but I think one or more of the dependencies we're using isn't yet compatible with node 14, so I've been leaving things at node 12.

I recommend downgrading your local setup to node 12 for this project until we can upgrade to 14. I don't think the version of npm should make much of a difference, but that's worth upgrading to the latest.

maitanavarro:
Sounds good! It was able to work once I switched to node 12. I think the only thing that looks off is the layering of boxes on the lower left hand corner
Screen Shot 2021-04-06 at 5 02 04 PM

maitanavarro (approved):

PR #79, opened 2021-04-05
Convert map dropdown to use Downshift

justinanderson:
Swaps out semantic-ui-react for a lighter-weight custom dropdown select menu based on [downshift](https://www.downshift-js.com) and [popper.js](https://popper.js.org).

Should be useful for implementing more a more accessible version of the hamburger menu and for implementing a geocoding search field with more flexible styling.


maitanavarro (commented):
I received this error when trying to run it, also once again not sure if this is an issue on my end
Screen Shot 2021-04-06 at 3 29 51 PM

justinanderson:
Try this again now that you're on node 12 and this has been rebased onto the changes from #78?

maitanavarro:
Just tried it again! Unfortunately, I still have the same error
Screen Shot 2021-04-07 at 6 48 46 PM

PR #80, opened 2021-04-09
Sidebar redesign [WIP]

jenniferwang99:
This PR:
- Moves the sidebar up the DOM hierarchy so it's a sibling of the map.
- Makes the sidebar collapsible and changes the position of floating elements on the map when the sidebar is visible.
- Persists the selected region when the map layer is changed
- Gets rid of the "Change flow direction" bubble

Some issues I am looking into are:
- the search bar is very much broken and not working
- the map does not actually resize
- the sidebar has no transition animation currently


src/css/sidebar.css
```diff
@@ -0,0 +1,57 @@
...
+.sidebar-content {
+    overflow-y: auto;
+    height: calc( 100% - 52px );
+}
+
+.sidebar-container.closed {
+    display: none;
+}
+
+.toggleButton {
```
justinanderson, line 21: Style SidebarButton to match [the Figma design](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1472%3A11484).

src/components/display.js
```diff
@@ -5,32 +5,30 @@ import ReactTooltip from 'react-tooltip';
 import Collapsible from "react-collapsible";
```
justinanderson, line 5: Rename display.js to sidebar.js to be more accurate.

src/css/sidebar.css
```diff
@@ -0,0 +1,57 @@
...
+.arrow {
+    border: solid black;
+    border-width: 0 3px 3px 0;
+    display: inline-block;
+    padding: 3px;
+    margin: auto;
+}
+
+.toggleButton.closed {
+    left: 0px !important;
```
justinanderson, line 46: We shouldn't be doing anything that requires putting an `!important` into CSS. Also, it doesn't seem to be necessary here.

src/css/sidebar.css
```diff
@@ -0,0 +1,57 @@
...
+
+.sidebar-container {
+    height: 100%;
+    background-color: white;
+    z-index: 2;
+}
+
+.sidebar-content {
+    overflow-y: auto;
+    height: calc( 100% - 52px );
```
justinanderson, line 14: We should need to do any CSS calculations. This wouldn't adjust to the height changes of its siblings. You could change this to `flex: 1 1 auto` and add `display: flex` and `flex-direction: column` to `.sidebar-container` to keep things the right height.

src/components/display.js
```diff
@@ -43,18 +41,24 @@ let Display = class Display extends React.Component {
       sa1_codes,
     } = this.props.select;
 
-    const { flowDirection, mapType } = this.props;
+    const { mapType } = this.props;
 
     const PanelContainer = (props) => (
-      <div className="flex-parent flex-parent--column flex-parent--space-between-main absolute top left w240 h-full pt60 pb36 mr12 z2">
+      <div className={`panel-container h-full`}>
```
justinanderson, line 47: `h-full` seems to be unnecessary.

src/components/legend.js
```diff
@@ -83,7 +86,7 @@ let Legend = class Legend extends React.Component {
     };
 
     return (
-      <div className="bg-white absolute bottom left ml12 mb36 py12 px12 shadow-darken10 round z1 w300">
+      <div className={`legend ${this.props.sidebarOpen ? "resize" : ""} bg-white absolute bottom left mb36 shadow-darken10 round z1 w300`}>
```
justinanderson, line 89: Rather than making multiple components subscribe to `sidebarOpen` state changes, it might be easier to have a higher level component like Main watch `sidebarOpen`'s value and toggle a class on its div. Then all of the other components could rely on one CSS rule to override their position, e.g.

```css
.main.sidebarOpen .legend, .main.sidebarOpen .toggleButton {
    left: 320px;
}
```

You'd need to give Main the .main class and give Legend actual CSS instead of relying on Assembly's classes, which are weirdly aggressive with !important.

justinanderson (commented):
You can remove semantic-ui.css from index.html. It looks like the only useful part of it for the map was
```css
body, html {
    height: 100%;
}
```
I recommend adding that to something high-level like App.css.

Also, Header should be adjusting its offset and width when the sidebar toggles, since it'll eventually get a centered [table/grid] segmented control. If you use the same CSS selector trick I suggested in a comment, the only other change you'd need to make to Header's `.container` is to remove `width: 100%` and add `left: 0; right:0` to make it size itself correctly.

public/index.html
```diff
@@ -12,10 +12,10 @@
       href="https://api.tiles.mapbox.com/mapbox-gl-js/v0.39.1/mapbox-gl.css"
       rel="stylesheet"
     /> -->
-    <link
+    <!-- <link
```
justinanderson, line 15: I'd suggest just deleting this rather than commenting it out.

justinanderson (changes requested):
One super tiny thing, then this is good for merging.

justinanderson (approved):

PR #82, opened 2021-04-12
Implement search field for new design using downshift-js

### LINKED ISSUE #83: Implement search field for new design
Search functionality
- [x] Make a custom search field element using downshift-js's useComboBox() hook.
- [x] Include a local list of SA2 names in search suggestions.
- [x] Include results from the Mapbox Geocoding API.
- [ ] Make the search field's highlight and selection affect the map's highlight and selection.

Appearance states
- [x] Floating
- [ ] Sidebar, editing or empty
- [ ] Sidebar, filled

Performance
- [x] Cache Geocoding API results to minimize redundant API calls.
- [ ] Load search field's GeoJSON only after map has finished loading.

justinanderson:
Will resolve #83 when finished.


justinanderson:
The search field and updated sidebar coexist, but not all that elegantly. I'd like to get this merged in so that its redux state changes are available for other developers to use, in particular `selectedFeature`, which includes all of the data needed for populating the sidebar when showing a single location.

Some outstanding issues which should be handled by future pull requests:

- The search field doesn't yet have any special behavior for drawing itself differently when the sidebar is open. It's currently always a rounded rectangle with a shadow.
- The search field is pretty inefficient at the moment. It rerenders itself when almost anything about it changes.

justinanderson:
Ok, everything coexists better now. It's still inefficient, but at least the sidebar and search field aren't fighting for screen space.

src/components/SearchField.js
```diff
@@ -0,0 +1,276 @@
...
+        return {
+          ...localFeature,
+          secondary: f.place_name,
+          relevance: f.relevance,
+        }
+      // Filter out results which aren't inside any of the regions this app
+      // tracks. For example, a search for "west lak" includes Lake Mundi in
+      // Victoria, even though the geocoding request was restricted to Sourth
+      // Australia.
+      }).filter(f => f.primary)
```
jenniferwang99, line 183: Not really sure if this really matters performance-wise but if you put this filter statement before the .map() you'll make fewer map calls (and .find() calls too) so could be good just in case

src/components/header.js
```diff
@@ -7,89 +7,79 @@ import { setHeaderOption } from "../redux/action-creators";
...
         </div>
       </div>
-    );
-  }
+    </div>
+  );
+}
+
+Header.propTypes = {
+    path: PropTypes.string.isRequired,
```
jenniferwang99, line 82: The path prop is no longer used in this file, could you remove it? The same would go for the PropTypes import and the mapStateToProps function in here since it looks like we no longer need it

jenniferwang99 (commented):
This looks really good! I found two pretty small/minor things but the search bar looks pretty good/functional.

Do you plan on doing anything with the inline css (since there's a decent amount of it) later? Honestly it would be great to move some of it into their own files now but everything's functional so it's up to you if you want to do it later.

src/components/SearchField.js
justinanderson: It's actually filtering out things that failed to map. It's not obvious thanks to spreading `...localFeature`, but the `primary` key it's filtering on doesn't exist until after the map has done its thing.

jenniferwang99 (approved):
Looks good to me!

PR #84, opened 2021-04-23
Update menu styling and add dim shield [WIP]

maitanavarro:


src/css/header.css
```diff
@@ -12,34 +12,79 @@
...
+    font-size: 1.1rem;
+    font-weight: bold;
+    display: flex;
+    flex-direction: column;
+    margin-left: 25px;
+    margin-top: 15px;
+    margin-bottom: 20px;
+}
+
+.close-icon{
```
justinanderson, line 53: Add `cursor: pointer;` so users know it's clickable.

src/css/header.css
```diff
@@ -12,34 +12,79 @@
...
+    box-shadow: 0 0 0 100vmax rgba(0,0,0,.3);
+}
+
+
 .dropdown-menu {
-    display: flex;
-    flex-direction: column;
     margin-left: 2rem;
     margin-right: 1rem;
+    width: 19%;
```
justinanderson, line 29: The width here seems to do nothing and can be removed.

justinanderson (commented):
In addition to the line comments Ieft, I'd prefer if font sizes, padding, and margins were done in px rather than rem, e.g. menu-title should be 18px rather than 1.1rem.

src/components/container.js
```diff
@@ -1,38 +1,96 @@
...
     title: PropTypes.string.isRequired,
     children: PropTypes.array,
   };
+
   render() {
     const { title, children } = this.props;
-    const container = {
+
+    const popUpBox = {
+      width: "100%",
```
justinanderson, line 15: Once the infinite recursion of Main and Routes is fixed, the modal windows do actually render, though they end up behind the map. To fix that, you'll need to set popUpBox's position to absolute (so z-index works on it) and its z-index to at least 3 so it's above the map and everything else.

After that, it's just a matter of getting the rest of popUpBox and container's styling right and giving it a way to be dismissed. If you set the popUpBox as width and height 100% and a translucent black background, you can use it as the clickable dim shield while container holds the actual modal content.

justinanderson (changes requested):

src/components/container.js
```diff
@@ -1,38 +1,160 @@
...
-        <div style={content}>
-          <h3 style={titleBox}>{title}</h3>
-          {children}
-        </div>
-      </div>
+
+      <div>
+        
+        {
+          this.state.showPopUp
```
justinanderson, line 104: `showPopUp` and `setHeaderOption` are both obsolete now that the visibility of modal dialogs is controlled solely by the URL. All use of those can be removed from container.js and header.js, simplifying them a fair bit.

justinanderson (changes requested):

justinanderson:
Oh, also there's a runtime error for each modal dialog because they expect a `modal` prop to come in from redux, but that's obsolete now. Any use of modal in about.js, research.js, methods.js, and faq.js should be removed.

justinanderson (approved):

justinanderson:
👍

PR #96, opened 2021-04-26
Backend dev [WIP] 

quocphan568:
Create the PR to review
It is still in progress and NOT ready to be merged yet


api/database/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM postgres:alpine
+COPY *.sql /docker-entrypoint-initdb.d/
+ADD scripts/1_init.sql /docker-entrypoint-initdb.d
```
justinanderson, line 3: These are the wrong paths for these scripts. It looks like all of the database import stuff was committed mid-move from a different folder.

api/database/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.3'
+
+services:
+  db:
+    build: ./db
+    container_name: postgres
+    ports:
+      - "5431:6666"
+    environment:
+      - POSTGRES_USER=SADashboard
```
justinanderson, line 10: This user and database need to be created in an entrypoint script if they're to be used by the Docker Compose setup.

api/database/2_copy.sql
```diff
@@ -0,0 +1 @@
+COPY table_name FROM '/data/datasource.csv' DELIMITER ',' CSV HEADER;
```
justinanderson, line 1: This won't work until you pass it a real table name.

api/database/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.3'
```
justinanderson, line 1: All of this should be part of the main docker-compose.yml file. The goal is to have dashboard, api, and db all running together.

api/database/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
...
+    build: ./db
+    container_name: postgres
+    ports:
+      - "5431:6666"
+    environment:
+      - POSTGRES_USER=SADashboard
+      - POSTGRES_PASSWORD=dashboard
+      - POSTGRES_DB=SAData
+    volumes:
+      - ./data:/data
```
justinanderson, line 14: This should reference the actual path where the CSV lives rather than to a local copy not in version control, i.e. `/data` should be `../../src/data/`. And the reference to the CSV in the entrypoint COPY script should use the correct file name.

justinanderson (changes requested):
After you address my line comments, your next steps should be to:
- Make the Docker setup for the database actually work and then merge that into the base docker-compose.yml file.
- Make the Flask server pull data from the Postgres database instead of printing out static "Hello" content. You'll need the Postgres server to be a part of the Docker Compose setup for that, and you'll need to use a library like [psycopg2](https://pypi.org/project/psycopg2/) to communicate with it.

api/Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM python:3
```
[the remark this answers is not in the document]
quocphan568: Fixed, learnt new stuff, thanks for your sharing

api/database/1_init.sql
```diff
@@ -0,0 +1,12 @@
+CREATE TABLE CSVDATA(ID INT PRIMARY KEY NOT NULL, SA2Code VARCHAR(1000) NOT NULL, X1 INT NOT NULL, X INT NOT NULL, StateName TEXT NOT NULL, STateCode INT NOT NULL, SA2Name16 TEXT NOT NULL, PersonsNum INT NOT NULL,
```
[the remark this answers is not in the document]
quocphan568: changed the column name,
set SA2_code is not Null only
Renamed database to SAData

api/database/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.3'
+
+services:
+  db:
+    build: ./db
+    container_name: postgres
+    ports:
+      - "5431:6666"
```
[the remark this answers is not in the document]
quocphan568: Fixed

api/database/docker-compose.yml
```diff
@@ -0,0 +1,14 @@
+version: '3.3'
+
+services:
+  db:
+    build: ./db
+    container_name: postgres
```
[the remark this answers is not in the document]
quocphan568: removed container_name

api/database/docker-compose.yml
quocphan568: updated to the main docker-compose.yml file

api/database/Dockerfile
quocphan568: fixed

api/Dockerfile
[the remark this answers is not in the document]
quocphan568: got error "Failed to build greenlet" with  3.9.4-alpine version , temporarily  change back to python:3

api/database/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM postgres:alpine
+WORKDIR /docker-entrypoint-initdb.d
```
justinanderson, line 2: Most of this is unnecessary:
- Setting the WORKDIR doesn't matter since all of the destination paths below it use absolute paths.
- The 2 ADDs repeat what COPY already did.
- The chmod seems to be unnecessary.
- postgres:alpine already exposes port 5432.

If we take all of that away, we're left with `COPY *.sql /docker-entrypoint-initdb.d/`, which is already being done via (and actually being overridden by) the Docker volume in docker-compose.yml `./api/database:/docker-entrypoint-initdb.d`. So this entire Dockerfile can be deleted, and the line `build: ./api/database` in docker-compose.yml can be replaced with `image: "postgres:alpine"`.

And ideally we'd move all of the .sql files in api/database/ into a separate api/database/docker-entrypoint-initdb.d/ folder so we're not sharing more files with the db container than we mean to.

api/wait-for-it.sh
```diff
@@ -0,0 +1,182 @@
+#!/usr/bin/env bash
```
justinanderson, line 1: Alpine Linux doesn't come with bash installed, so wait-for-it.sh dies immediately with `env: can't execute 'bash': No such file or directory`. wait-for-it uses statement which require bash, so I recommend switching to one of the sh-compatible options from https://docs.docker.com/compose/startup-order/ like https://github.com/Eficode/wait-for.

justinanderson, line 1: The api container fails to start because wait-for-it.sh's file permissions don't include execution. You'll need to push a file permission change on the file.

justinanderson (changes requested):
In addition to the changes requested in line comments, the output of the API isn't right. It's not in GeoJSON format and doesn't include the geometry data for each feature, which still needs to be imported into the database from a separate file.

api/database/1_init.sql
```diff
@@ -0,0 +1,70 @@
+CREATE TABLE SAData(
+    id INT PRIMARY KEY NOT NULL, 
+    sa2_code VARCHAR(1000) NOT NULL ,
```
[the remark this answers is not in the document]
quocphan568: :) thanks Justin, I got this error before. Updated

api/Dockerfile
```diff
@@ -0,0 +1,15 @@
...
+
+WORKDIR /app
+
+ENV FLASK_APP=app.py
+
+ENV FLASK_ENV=development
+
+COPY ./requirements.txt .
+
+RUN pip install -r requirements.txt
```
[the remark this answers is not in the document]
quocphan568: I did not get this error yesterday but get it today. Thanks Justin. Updated

api/app.py
```diff
@@ -0,0 +1,31 @@
...
+  return {
+    "field1": "Hello",
+    "field2": "From Backend"
+  }
+
+@app.route('/api/features.geojson', methods=['GET'])    
+def get_sa_data():
+  cur.execute('SELECT * FROM sadata')
+  rows = cur.fetchall()
+  print(rows)
```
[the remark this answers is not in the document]
quocphan568: Yes. Removed

api/database/docker-compose.yml
justinanderson: Just realized my comment was incorrect. The Postgres image is configured to watch for these environment variables when it configures itself.

api/database/Dockerfile
quocphan568: removed unnecessary code. updated docker compose
have not moved " all of the .sql files in api/database/ into a separate api/database/docker-entrypoint-initdb.d/ folder" yet.

api/wait-for-it.sh
quocphan568: I did not get the error yesterday, but get it today :) thanks Justin
updated

quocphan568: changed wait-for-it.sh' to wait-for. Tested on my local

api/database/2_copy.sql
```diff
@@ -0,0 +1,234 @@
+-- \copy sadata FROM '/src/data/SA2_info_for_dashboard.csv' with DELIMITER ',' CSV HEADER;
+
+INSERT INTO SAdata (
```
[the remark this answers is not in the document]
quocphan568: updated. tested

api/database/Dockerfile
justinanderson: Per the middle of my comment, you can actually completely delete api/database/Dockerfile if you change one line of docker-compose.yml.

api/wait-for-it.sh
justinanderson: I still get a permission error here: `sh: ./wait-for: Permission denied`. On my computer, the wait-for script isn't executable. This should show `-rwxr-xr-x`:
```sh
$ ls -al api/wait-for 
-rw-r--r--  1 jander  staff  4059 May  3 18:59 api/wait-for
```
I'm guessing you're using Windows? I know on a Mac or Linux machine I'd fix this with `chmod +x api/wait-for`, but I'm not familiar enough with Windows file permissions to suggest a solution off the top of my head.

quocphan568: Yes, I am using Windows.
I will update the permissions

quocphan568: added permissions for the file

api/app.py
```diff
@@ -0,0 +1,100 @@
...
+# insert data from SA_SA2s.geojson file to database
+
+def insertDataToDatabase():      
+  is_database_has_data =  cur.execute('SELECT EXISTS (SELECT 1 FROM sa2data)')
+  
+  if(is_database_has_data is None):
+    with open('/data/SA_SA2s.geojson','r') as file:
+      df = json.load(file)
+
+      for feature in tqdm(df['features']):
```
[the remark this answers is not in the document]
quocphan568: removed, thanks Justin

api/database/2_copy.sql
```diff
@@ -0,0 +1,5 @@
+
+IF(NOT EXISTS(SELECT 1 FROM SADatabase.sadata))    --only add data if sadata database is empty
```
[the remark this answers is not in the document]
quocphan568: removed, thanks Justin

docker-compose.yml
```diff
@@ -1,7 +1,34 @@
...
+      - "5000:5000"
+    volumes:
+      - ./src/data/SA_SA2s.geojson:/data/SA_SA2s.geojson
+      - ./api:/app
+    environment:
+      FLASK_ENV: development
+    depends_on:
+      - db 
+    command: sh -c './wait-for db:5432 -- python app.py'
+  # dashboard:
```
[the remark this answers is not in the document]
quocphan568: sorry, my bad, I tested it on my local and forget to uncomment it.

api/app.py
```diff
@@ -0,0 +1,100 @@
...
+        host="db",
+        port="5432")
+cur = con.cursor()
+
+# insert data from SA_SA2s.geojson file to database
+
+def insertDataToDatabase():      
+  is_database_has_data =  cur.execute('SELECT EXISTS (SELECT 1 FROM sa2data)')
+  
+  if(is_database_has_data is None):
```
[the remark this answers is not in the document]
quocphan568: thanks Justin, It is more clean

api/app.py
```diff
@@ -0,0 +1,100 @@
+from flask import Flask, request, jsonify
```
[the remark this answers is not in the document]
quocphan568: formated, thanks Justin

api/app.py
```diff
@@ -0,0 +1,31 @@
...
+import psycopg2
+
+app = Flask(__name__)
+
+conn = psycopg2.connect(database="SADatabase",
+        user="user",
```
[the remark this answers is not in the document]
quocphan568: added .env file which keep all environment variables and updated docker-compose file

api/app.py
```diff
@@ -0,0 +1,206 @@
...
+    command += " j.SA4_NAME16, j.GCC_CODE16, j.GCC_NAME16, j.STE_CODE16, j.STE_NAME16, j.AREASQKM16,"
+    command += " persons_num, median_persons_age, males_num, median_male_age, females_num, median_female_age,"
+    command += " percentage_person_aged_0_14, percentage_person_aged_15_64, percentage_person_aged_65_plus, earners_persons,"
+    command += " median_age_of_earners_years, gini_coefficient_no, median_aud, mean_aud, income_aud, highest_quartile_pc,"
+    command += " income_share_top_5pc, third_quartile_pc, lowest_quartile_pc, income_share_top_1pc, second_quartile_pc,"
+    command += " income_share_top_10pc,popfraction,quartile,occup_diversity,bsns_entries,bsns_exits,bsns_growth_rate,bridge_diversity,"
+    command += " bridge_rank1,bridge_rank2,bridge_rank3,bridge_rank4,bridge_rank5,time_q1,time_q2,time_q3,time_q4,tot_time,"
+    command += " fq1,fq2,fq3,fq4,income_diversity,raw_inequality,inequality,inflow_r1,inflow_r2,inflow_r3,outflow_r1,outflow_r2,outflow_r3,"
+    command += " spent_r1,spent_r2,spent_r3,gain_r1,gain_r2,gain_r3,exchanged_r1,exchanged_r2,exchanged_r3, j.geometry"
+    command += " FROM sadata c INNER JOIN sa2data j ON c.SA2_code = j.SA2_MAIN16"
```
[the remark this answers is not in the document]
quocphan568: yes, It looks not good. I am still looking for a way to do it. There are some columns we do not need in sadata, but seems like there is no way to "Exclude" these columns by SQL query

api/app.py
```diff
@@ -0,0 +1,100 @@
...
+ 
+  cur.execute('SELECT * FROM sa2data')
+  rows = cur.fetchall()
+  
+  json_format =[]
+  for row in rows:
+    json_format.append(row)
+
+
+  geojson_data = {
```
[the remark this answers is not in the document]
quocphan568: fixed, thanks Justin

api/database/Dockerfile
quocphan568: updated. 
Removed Dockerfile of the database. 
Moved all of the .sql files in api/database/ into api/database/docker-entrypoint-initdb.d/ folder

justinanderson:
Obsolete with Atlas v2 changes.

PR #97, opened 2021-04-29
Welcome modal

Tibblist:


justinanderson (changes requested):
- Please refrain from committing lines that are just style changes. It's fine to leave them in the branch now that they're there, but they aren't relevant to the PR and make it harder for reviewers to discern which lines are meaningful. I know some people use editors which enforce a particular coding style on save, but this project doesn't have a defined coding style, so that ends up adding noise to commits.

- The rest of these comments are mostly CSS changes. In case you didn't know, Figma provides the CSS needed to replicate an object's appearance. For whichever object you've got selected, select "Inspect" in the righthand pane up where it says "Design", "Prototype", and "Inspect", then scroll down to CSS. It's often not optimal CSS, but it can make the specifications clearer when you aren't intimately familiar with all of the fiddly bits in the mockup.

- The welcome box's content should match the Figma mockup.
    - Its padding should be 30px 35px, not 40px all around.
    - Text color should be #333333, not black.
    - The title should be larger and bold.
    - The body should be larger, have a greater line height, and be aligned left rather than justified.
    - No "Get started below!" or bolded organization names.
    - No "Explore the Dashboard" or "Take a Tour" buttons.
    - The close button should be 15px from the right of the box, not 10px.
    - The close button shouldn't scroll with then rest of the welcome box's content when it overflows.

- The welcome box itself should be positioned, sized, and styled as in the Figma mockup.
    - It should be part of the flex layout of its parent `.sidebar-container` instead of opting out with `position: absolute`. This will help resolve some of the other notes in this review.
    - Its top edge should always be 10px from the bottom of the search field. Note the search field already has a 5px bottom margin.
    - Its border-radius should be 5px, not 8px.
    - Its box-shadow should be the same as `.searchField`'s.
    - Its height should attempt to fit its content but never grow so much as to overlap the legend. I expect this should be straightforward to do with flex-grow on the welcome box and a dummy DOM sibling of the welcome box set to the same height as the legend.

- The welcome box shouldn't block interaction with the search field or the layer select dropdown in the legend. I think this is resolved by not using absolute positioning for it per above.

src/components/sidebar.js
```diff
@@ -112,15 +122,17 @@ let Sidebar = class Sidebar extends React.Component {
...
         <div className={`sidebar-container`}>
-            <SASearchField />
-          {this.props.selectedFeature ?
-            <TopPanel /> :
-            ""
-          }
+          <SASearchField />
+          {this.props.selectedFeature ? <TopPanel /> : <>
+            <Modal />
+            <Legend/>
```
justinanderson, line 134: Legend should stay in Main, as moving Legend out of Main and into .sidebar-container means it disappears when there's a selection.

The first solution that comes to mind for keeping Modal from visually overlapping Legend while Legend is in Main is to place an empty <div /> in .sidebar-container with the same height as Legend and to make sure Legend's z-index is in front of .sidebar-container.

Tibblist: I can make the legend stay rendered when a feature is selected if we want, I intentionally made it disappear since I thought that was the goal behavior. If it needs to be absolutely positioned outside then just using margin at the bottom that is as large as the legend should do it.

justinanderson: Good idea. A bottom margin on .sidebar-container when there's no selected feature would be cleaner.

The desired behavior for the legend is in the first few screen of "v4 / Clickable Prototype" in Figma:
- [No selection, welcome visible](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15227)
- [No selection, welcome dismissed](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15234)
- [Selection, sidebar not collapsed](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15279)
- [Selection, sidebar collapsed](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15279)

Tibblist: This is a potentially best of both worlds solution, it now renders as it did before when sidebar is opened/collapsed. At the same time it renders in such a way when no feature is selected that if we resize it down the line the rest of the sidebar will auto-adjust and we won't have to remember to change the bottom margin on the sidebar

justinanderson: That works for me if you can get the welcome box to clip/overflow: scroll its contents when the window height is short instead of letting the Legend fall off the page.

Also, some stuff I noticed:
- There should be 5px more space between the search field and welcome box.
- Legend will need its own `pointer-events: auto` as being in the floating sidebar means it ignores clicks by default.
- The welcome box is on top of the search field's suggestions. The welcome box doesn't seem to actually need its z-index definition.

Tibblist: Fixes applied, I'm used to working with ui frameworks like material ui or ant to handle flexbox aspects so it took a bit to manually control some aspects of the flexbox using actual div elements but it should function properly.

justinanderson: Yeah, flexbox takes some getting used to, but it's pretty powerful. I had a few other comments, but it took as long to make the changes as it would have to write them up. Could you review the extra commits I've added?

Tibblist: They look good, only question is about bringing back modal.js. Didn't seem like it was very generic or re-usable so to not confuse people down the line into thinking it's a general purpose modal component maybe we should move it to a deprecated folder or leave it in a branch we can reference later if we need it.

justinanderson: If Maita's unpushed work on the hamburger menu implements modal dialogs the way I imagine she will, it'll be slightly more complicated for her to rebase her work onto master if modal.js disappears. If she doesn't end up using it, I'll ask her to delete it as part of the PR for her work.

justinanderson:
Rebased to get rid of package-lock.json commit which was preventing a rebase and merge.

PR #99, opened 2021-04-30
Sidebar Styling

jenniferwang99:
Updated sidebar styling to match the Figma design.

I think I just don't know how to export svgs from Figma but I tried to export the comparison button like three different ways and still didn't manage to get it to show up right so any pointers on how to correctly export that would be appreciated thank you


src/assets/compare.svg
```diff
@@ -0,0 +1,6 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
justinanderson, line 1: I didn't see anything weird with Compare myself, but Favorite would occasionally be a jumble of black and white.

I did Object > Outline Stroke to the Compare and Favorite buttons in Figma, and that seems to have simplified their geometry. If you export them now, they should make much smaller SVGs and not render weirdly.

justinanderson (changes requested):

justinanderson (approved):

PR #101, opened 2021-05-05
Initial UI testing

Tibblist:


package.json
```diff
@@ -54,6 +56,33 @@
...
+    "@babel/preset-react": "^7.13.13",
+    "@testing-library/jest-dom": "^5.12.0",
+    "babel-jest": "^26.6.3",
+    "enzyme": "^3.11.0",
+    "enzyme-adapter-react-16": "^1.15.6",
+    "jest": "^26.6.0",
+    "jest-fetch-mock": "^3.0.3",
+    "jest-svg-transformer": "^1.0.0",
+    "jest-transform-css": "^2.1.0",
+    "react-test-renderer": "^17.0.2",
```
justinanderson, line 70: Some of these devDependencies are already installed via react-scripts with specific compatible versions. I'd prefer to follow react-scripts' recommendation to not override any of its dependencies.

We can drop `@babel/core`, `@babel/present-env`, `@babel/preset-react`, `babel-jest`, and `jest`.

Also, the version of `react-test-renderer` should be `^16.14.0` since we're not using React 17.

src/redux/reducer.js
```diff
@@ -106,6 +108,11 @@ function fetchFeatures() {
 
 function loadFeatures() {
   return function (dispatch) {
+    if (IS_TEST) {
+      dispatch({type: "FEATURES", payload: JSON.parse(loadGeoJSON("./src/data/" + geojsonURL)).features}); 
+      return;
+    }
+
```
justinanderson, line 115: Shouldn't this not be necessary if the mock is working?

src/tests/Map.test.js
```diff
@@ -0,0 +1,20 @@
...
+import { loadFeatures } from "../redux/reducer";
+
+beforeEach(() => {
+  resetState()
+  fetchMock.mockIf("SA_dashboard.geojson", loadGeoJSON("./src/data/SA_dashboard.geojson"))
+  store.dispatch(loadFeatures())
+})
+it("renders correctly", () => {
+  // eslint-disable-next-line no-undef
+  fetchMock.mockResponse(loadGeoJSON("./src/data/SA_dashboard.geojson"));
```
justinanderson, line 17: Isn't this already covered by the .mockIf() in beforeEach()?

src/utils/constants.js
```diff
@@ -0,0 +1,2 @@
+/* eslint-disable no-undef */
+export const IS_TEST = process.env.NODE_ENV === "test";
```
justinanderson, line 2: There's an existing constants file in /src/constants.js. It might be better put `IS_TEST` in there. There might be an argument for keeping test-specific constants in a separate file, but `IS_TEST` is one variable that's never going to be needed by test files.

justinanderson (commented):
`npm test` reports the tests pass on my machine, but it produces a lot of warnings like `Warning: An update to ConnectFunction inside a test was not wrapped in act(...).` and `downshift: The ref prop "ref" from getInputProps was not applied correctly on your element`. Are those expected or specific to my machine?

src/redux/reducer.js
Tibblist: Yeah the rebase got a bit messy and this must have slipped in

PR #102, opened 2021-05-07
Refactor sidebar, resolve #85

### LINKED ISSUE #85: Refactor sidebar contents
- [x] Refactor TopPanel into its own component and give it a more accurate name.
- [x] Use `selectedFeature.properties` from sidebar's state instead of `select` to populate the location details in TopPanel.
- [x] Populate TopPanel's content more dynamically
    - [ ] Separate components for generating each level: Collapsible section header, single metric, etc.
    - [x] Add a lookup table and function for naming collapsible sections and mapping the various properties from `selectedFeature.properties` to them e.g.
    ```js
    const properties = this.props.selectedFeature.properties

    const propertiesMapping = [
      {
        id: "demographics",
        title: "Demographics",
        contents: [
          {
            id: "persons_num",
            title: "Population",
          },
          {
            id: "median_aud",
            title: "Median Income",
          }
        ]
      }
    ]
    
    propertiesMapping.map((section) => (
      <Collapsible trigger={section.title}>
        {section.contents.map((metric) => (
          <Metric title={metric.title} value={properties[metric.id]} />
        ))}
      </Collapsible>
    ))
    ```

gerardott:


src/components/LocationDetails.js
```diff
@@ -0,0 +1,82 @@
...
+          {section.content.map((metric) => (
+            renderMetric(metric)
+          ))}
+        </Collapsible>
+      ))}
+    </div>
+  )
+}
+
+LocationDetails.propsTypes = {
```
justinanderson, line 78: Typo here. Extra 's' in "propTypes".

src/components/LocationDetails.js
```diff
@@ -0,0 +1,82 @@
...
+    {
+      title: 'Growth Summary',
+      content: [
+        { id: 'income_diversity', label: 'GDP Growth Potential', desc: 'Economic growth is an increase in the production of economic goods and services,compared from one period of time to another... Traditionally, aggregate economic growth is measured in terms of gross national product (GNP) or gross domestic product (GDP), although alternative metrics are sometimes used.' },
+        { id: 'bridge_diversity', label: 'Job Resilience', desc: 'The ability to adjust to career change as it happens and,by extension, adapt to what the market demands.' },
+        { id: 'bsns_growth_rate', label: 'Business Growth Index', desc: 'The growth rate is the measure of a company’s increase in revenue and potential to expand over a set period.' },
+        { id: 'SA1_7DIGITCODE_LIST', label: 'Included SA1 Regions' },
+      ]
+    }
+  ];
```
justinanderson, line 34: I'd like to see the propsMapping object split out into a separate file to make it easier for our data folks to update. Also, it will quickly become large and clutter this component if it remains here.

src/components/LocationDetails.js
```diff
@@ -0,0 +1,82 @@
...
+
+function LocationDetails(props) {
+  const featureProps = props.feature.properties;
+
+  const propsMapping = [
+    {
+      title: 'Demographic Summary',
+      content: [
+        { id: 'persons_num', label: 'Population' },
+        { id: 'median_aud', label: 'Median Income', format: (val) => `AUD ${val}` },
```
justinanderson, line 14: I'd prefer to stick with toLocaleString() for formatting currency as is currently done for [sa2_properties in map.js](https://github.com/CxSci/SA-dashboard/blob/master/src/components/map.js#L401-L404), as this project may be adapted in the future for other countries and other currencies, and it'll be easier if we don't hardcode formats.

src/components/LocationDetails.js
```diff
@@ -0,0 +1,82 @@
...
+        { id: 'bsns_growth_rate', label: 'Business Growth Index', desc: 'The growth rate is the measure of a company’s increase in revenue and potential to expand over a set period.' },
+        { id: 'SA1_7DIGITCODE_LIST', label: 'Included SA1 Regions' },
+      ]
+    }
+  ];
+
+  const renderMetric = (metric) => {
+    let value = featureProps[metric.id];
+    if (typeof value === 'number') {
+      value = value > 10 ? Math.floor(value) : Math.floor(value * 10000) / 10000;
```
justinanderson, line 39: Why floor all numbers?

Also, it's possible that some values which should be formatted as strings will be stored as numbers and vice versa. For instance, `"SA2_MAIN16": 401011001` should not run through any number formatters.

It might be scale better to have propsMapping's `format` be an enum/key for a switch statement here, so we can be explicit about every property's format without needing to duplicate formatting functions, e.g.
```js
const propsMapping = [
    ...
    { id: 'median_aud', label: 'Median Income', format: 'currency_aud' },
    ...
    { id: 'quartile', label: 'Income Quartile', format: 'number' },
    { id: 'inequality', label: 'Inequality (lower is better)' },
    ...
]
...
const renderMetric = (metric) => {
    ...
    switch (metric.format) {
        case 'number':
            // format value as number
        case 'currency_aud':
            // format as AUD
        default:
            // leave as raw value by default
    }
    ....
}
```

justinanderson (changes requested):
A good start. Just needs to include all properties from selectedFeature and resolve the line comments I've made.

src/components/LocationDetails.js
```diff
@@ -0,0 +1,71 @@
...
+      </div>
+    )
+  }
+
+  return (
+    <div
+      style={{ overflowY: "auto" }}
+      className={`sidebar-content`}
+    >
+      {propsMapping.map((section) => (
```
justinanderson, line 54: Could you make LocationDetails render `{props.children}` here, so it's possible to pass other collapsible sections in to prepend this overflowing div? The "Locations to Compare" work Ethan's doing in #104 will need to be passed along into there, as well as other components in the future. So instead of what Ethan's currently got in sidebar.js, https://github.com/CxSci/SA-dashboard/blob/203dabfd575d667aaed0dfe87f32bb9744c81ff7/src/components/sidebar.js#L67-L69, he'd do:
```js
<LocationDetails feature={this.props.selectedFeature}>
  <LocationCompare />
</LocationDetails>
```

justinanderson (changes requested):

PR #104, opened 2021-05-10
Locations to compare

Tibblist:


src/components/LocationCompare.js
```diff
@@ -0,0 +1,92 @@
...
+    "width": "161px",
+    "height": "36px",
+    "left": "79.5px",
+    "top": "0px",
+    "background": "#FFFFFF",
+    "border": "2px solid #000000",
+    "boxSizing": "border-box",
+    "borderRadius": "100px",
+    margin: "auto",
+    marginTop: 15,
```
justinanderson, line 36: The "Show Comparison" button should be more reusable. It's a kind of button we'll use more in future designs. Instead of using Figma's computed CSS sizing for this one example, let's use CSS matching the element's auto layout settings in Figma. So, instead of a static width, height, left, top, or setting a horizontal margin, rely on flexbox and internal padding to position and size the button. That should just take `alignSelf: "center"` and `padding: "10px 15px"` instead of the existing positioning and sizing code.

justinanderson (changes requested):

src/components/LocationCompare.js
Tibblist: Left it with figma defaults so it can be re-used as is anywhere, but allowed a custom style prop to enable any of the properties to be overridden at will if it needs to change alignment or sizing etc for a specific use case.

justinanderson: The goal is to have a reusable button component in the Show Comparison button's style which adapts to its content rather than a component with a predefined title and overridable dimensions. We can deal with that as a separate issue.

justinanderson (approved):

justinanderson:
@Tibblist In the future, please keep branches up to date with master by rebasing instead of merging. It keeps our commit history cleaner.

PR #107, opened 2021-05-12
Go comparison mode

gerardott:


src/css/sidebar.css
```diff
@@ -75,4 +75,29 @@
 
 .icon {
     margin-right: 12px;
+}
+
+.comparison-title {
+    display: flex;
```
justinanderson, line 81: This text should be black.

Padding should be 19px on top, 13px on bottom.

`align-items: stretch` appears to be unnecessary.

src/components/ComparisonSidebar.js
```diff
@@ -0,0 +1,51 @@
...
+        <SidebarButton />
+        <div className={`sidebar-container`}>
+          <div className="sidebarHeader">
+            <div className="comparison-title">
+              <Link to="/" style={{lineHeight: 0}}>
+                <BackIcon />
+              </Link>
+              <span style={{marginLeft: 10}}>
+                Comparing Locations
+              </span>
```
justinanderson, line 39: The span here is unnecessary. Its marginLeft could be a marginRight on the Link instead. Also, that margin should be 11, and lineHeight: 0 appears to do nothing.

If you set `display: flex` and `align-items: center` on the link, the back icon will vertically center itself correctly.

Also, I'd prefer we keep this CSS in sidebar.css rather than changing it up here with inline styles.

src/routes/index.js
```diff
@@ -7,12 +7,14 @@ import About from "../pages/about";
...
 import FAQ from "../pages/faq";
 import Methods from "../pages/methods";
+import Comparison from "../pages/comparison";
 
 const Routes = class Routes extends Component {
   render() {
     return (
       <Switch>
         <Route exact path="/" component={Main} />
+        <Route exact path="/comparison" component={Comparison} />
```
justinanderson, line 17: This conflicts with the work in #84. routes/index.js and Main will [swap positions](https://github.com/CxSci/SA-dashboard/pull/84/files#diff-3d74dddefb6e35fbffe3c76ec0712d5c416352d9449e2fcc8210a9dee57dff67L18-R21) in the component hierarchy, so routes/index.js will end up being just for components which need to render in addition to Main rather than instead of it.

Additionally, given my comment about ComparisonSidebar.js, you'll want /comparison routing to happen inside of sidebar.js anyhow.

justinanderson (changes requested):

src/components/ComparisonSidebar.js
gerardott: The  lineHeight: 0 fix a issue with Link that add some px to the BackIcon height, and that make arrow unaligned.

src/components/LocationToCompare.js
```diff
@@ -55,12 +59,20 @@ const LocationCompare = () => {
         );
       })}
       </div>
-      <ComparisonButton onClick={() => {}} />
-      <div>
-        <p style={disclaimerText}>Add up to 4 regions.</p>
-      </div>
+      {showButton && (
+        <>
+          <ComparisonButton onClick={() => { history.push('/comparison') }} />
```
justinanderson, line 64: I'd rather we use a react-router-dom <Link to="/comparison"> here instead of history.push(), as we've been using Link everywhere else so far.

The onClick prop of ComparisonButton can be made optional.

src/components/LocationToCompare.js
```diff
@@ -36,14 +38,16 @@ const disclaimerText = {
...
+  const history = useHistory();
+
   return (
     <div style={root}>
     <div style={listRoot}>
       {comparisonFeatures.map((feature) => {
         return (
-          <div style={featureRoot} key={feature.primary}>
+          <div style={featureRoot} key={feature.properties.SA2_MAIN16}>
             <div>{feature.primary || feature.properties["SA2_NAME16"]}</div>
```
justinanderson, line 51: This can be simplified to `{feature.properties["SA2_NAME16"]}`. The enclosing div and check for feature.primary are unnecessary.

src/components/ComparisonMode.js
```diff
@@ -0,0 +1,34 @@
+import React from "react";
+
+import { ReactComponent as BackIcon} from "../assets/search-icons/back.svg";
+import "../css/collapsible.css";
+import "../css/sidebar.css";
+import { Link } from "react-router-dom";
+import Collapsible from "react-collapsible";
+import LocationCompare from "./LocationToCompare";
+
+class ComparisonMode extends React.Component {
```
justinanderson, line 10: This component could use a more specific name than ComparisonMode. ComparisonMode could apply to any of the components we'll make for that mode. It's the comparison mode version of sidebar content, so potentially ComparisonSidebarContent.

justinanderson (changes requested):
There's a crash that can happen after leaving comparison mode.

Steps to reproduce:

1. Select a location
2. Click "Add to Comparison"
3. Click "Show Comparison"
4. Click back button to leave comparison mode
5. Click somewhere on the map outside of any region, like the sea

After doing that, I see this error twice, happening in both sidebar.js:42 and map.js:332.
```
TypeError: null is not an object (evaluating 'this.props.selectedFeature.properties')
```

There might be something about removing and readding the map which messes with the redux state for selectedFeature.

src/components/ComparisonSidebar.js
justinanderson: What about my other comments?

justinanderson (approved):

PR #110, opened 2021-05-12
Keep the map still when selecting a location 

### LINKED ISSUE #109: Keep the map still when selecting a location
- [ ] The map's initial bounds aren't centered now that we have a welcome dialog.
- [ ] The map jumps around under the mouse when a region is selected or deselected. It'll be a jarring experience for users if their selection jumps away from their cursor. The top right corner of the map's bounds should be pinned to keep it in place when it resizes for the sidebar.

justinanderson:
Pin northeast corner of map when resizing.

Take WelcomeDialog into account for initial map padding.

Resolves #109.

### DIFF
--- a/src/components/map.js
+++ b/src/components/map.js
@@ -67,7 +67,9 @@ let Map = class Map extends React.Component {
         [129, -38],
         [141, -26]
       ],
-      fitBoundsOptions: { padding: 70 },
+      // 70px padding around initial viewport, with an extra 310 on the left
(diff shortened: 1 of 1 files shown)


Tibblist (approved):
Looks solid, might be a good idea to make the length of the welcome dialog an exported variable that is imported here and to size it so that if we resize it down the line we won't have to check all the places that might need to further be adjusted.

PR #116, opened 2021-05-14
[WIP] Branding Bar

jenniferwang99:
Added footer bar at the bottom of main page displaying the branding for affiliated companies.

Color of footer bar can be toggled from a white background to a black background with `inDarkMode` prop.

WIP - sidebar acts funny when expanded


src/pages/main.js
```diff
@@ -43,6 +43,7 @@ const Main = (props) => {
           {!comparisonMode && <Map />}
         </div>
         {props.selectedFeature && !comparisonMode && <Legend absolute />}
+        <Footer inDarkMode/>
```
justinanderson, line 46: If you make `inDarkMode={!comparisonMode}`, it should update at the right times.

justinanderson (changes requested):
The sidebar pushes the branding bar off the bottom of the page is the screen is short, but you can fix that by changing the .sidebar-content's flex property to `flex: 1 0 0;`. That sets its flex-basis to 0, which means it starts at 0 height and only grows to fill the available space. The existing `flex: 1 0 auto` lets it fit its content first, which grows the whole sidebar and all of its ancestors now that .main uses flexbox.

Also, I'm seeing some weird SVG drawing errors in Safari. Sometimes only half of a logo is drawn. Firefox and Chrome are fine. I'll keep poking at it to see if I can tell what's going on.

justinanderson (approved):
This is all good.

I figured out the Safari bug. It turns out [Safari has spotty support for SVG's clip-path element](https://stackoverflow.com/a/41860519/85337) and seemingly randomly clips out parts of SVGs. Luckily, none of the branding SVGs actually need their clip-paths, they just got them as part of the export process. I'll fix those and add them to your branch before merging it.

Issue #114, opened 2021-05-14
Show inline comparison in sidebar when Locations to Compare is not empty


justinanderson:
When there is at least 1 location in "Locations to Compare" (1 feature in `comparisonFeatures`) and we're not in comparison mode (the "Show Comparison" button hasn't been clicked), the sidebar should show an inline comparison of the Locations to Compare and the currently selected location.

This is a lot like `<LocationDetails />`, but showing multiple locations per metric. So while a sidebar would look like [this](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1479%3A10932) while `comparisonFeatures` is empty, it would look like [this](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1479%3A12212) while `comparisonFeatures` is populated.

Note the locations to show details for in the sidebar includes both the contents of `comparisonFeatures` and whatever `selectedFeature` is if it's not already in `comparisonFeatures`.

justinanderson:
Resolved by #118.

PR #118, opened 2021-05-17
Add inline comparison for each metric

gerardott:
Closes issue #114.

### DIFF
--- a/src/css/collapsible.css
+++ b/src/css/collapsible.css
@@ -27,7 +27,9 @@ h2 {
 .Collapsible__contentInner p {
   margin-bottom: 10px;
   font-size: 14px;
-  line-height: 20px; 
+  line-height: 20px;
+  display: flex;
+  justify-content: space-between;
 }
 
 .Collapsible__contentInner p:last-child {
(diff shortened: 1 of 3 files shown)


src/css/collapsible.css
```diff
@@ -27,7 +27,9 @@ h2 {
 .Collapsible__contentInner p {
   margin-bottom: 10px;
   font-size: 14px;
-  line-height: 20px; 
+  line-height: 20px;
+  display: flex;
+  justify-content: space-between;
```
justinanderson, line 32: This affects the display of the Locations to Compare section too. I'd like to see a more specific CSS selector for these lines and a CSS class specific to groups of metrics, like `className={comparisonFts.length ? "comparison" : ""}` on line 18 of LocationDetails.js and then break these 3 lines into `.Collapsible__contentInner .comparison>p { ... }`

justinanderson (changes requested):

src/css/collapsible.css
gerardott: Not really, because in Locations to Compare section they used a `<div>` and not a `<p>` so it's not affected. 
But I'll add a class to `<p>` when is comparing.

justinanderson: The "Add up to 4 regions." text uses a p tag. Its center alignment was getting overridden to left alignment.

PR #119, opened 2021-05-20
Mini map

Tibblist:


justinanderson (changes requested):
In addition to line comments, the bit from the last review about regions needing stronger outlines is still there, e.g. 
compare [current state](https://user-images.githubusercontent.com/244952/119173091-bf91c800-ba34-11eb-9a7a-041e1591ddbb.png) to [the Figma design](https://user-images.githubusercontent.com/244952/119173147-d89a7900-ba34-11eb-97ca-f31bba86927c.png).

src/components/map.js
```diff
@@ -256,25 +263,23 @@ let Map = class Map extends React.Component {
...
-            1,
-            ["boolean", ["feature-state", "hover"], false],
-            1,
-            0.8,
-          ],
+          "fill-opacity": 0.8,
+        },
+      });
+      this.map.addLayer({
+        id: "sa2-borders",
```
justinanderson, line 274: `Error: Layer with id "sa2-borders" already exists on this map`

I actually don't get any borders on the minimap unless I rename this, because the borders from rest of the features already use that layer id. You can change sa2-borders here to anything else though and it works.

src/components/map.js
```diff
@@ -256,25 +263,23 @@ let Map = class Map extends React.Component {
...
+        type: "line",
+        source: "sa2-comp",
+        sourceLayer: "original",
+        layout: {},
+        paint: {
+          "line-color": "#000080",
+          "line-width": 2,
+          "line-opacity": 1.5,
```
justinanderson, line 282: I get `Error: layers.mini-borders.paint.line-opacity: 1.5 is greater than the maximum value 1`. 1.0 would match the Figma mockup. I don't know why it's 1.5 elsewhere in the app and how that doesn't produce errors in the console.

Also, `line-color` should be `#ffffff`.

justinanderson (changes requested):
Small bits. Almost done.

justinanderson (approved):

PR #120, opened 2021-05-20
Table view

### LINKED ISSUE #115: Show comparison table view when in comparison mode
[The table view](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15356) is the default view when in comparison mode.

We haven't implemented any special formatters / visualizations for metrics yet, so this task is limited to implementing the table that will eventually hold those visualizations. For this task, it's enough to display the same text values as the sidebar does in `<LocationDetails />`. Don't worry about data filtering or drawing line graphs.

Things to do:
- [ ] Same `<Collapsible />` section headers as the sidebar, just styled to be bigger
- [ ] Draw the rows with an alternating background color
- [ ] Fix the label column to 170px wide
- [ ] Fix each location in a 120px wide column
- [ ] Separate the columns by 25px

gerardott:
Closes #115


justinanderson (changes requested):
- Table columns should be narrower, per my earlier spec. We're designing for a smaller minimum screen size. In later development we can work our way to up sizing columns responsively, but for now I want to focus on the minimum.

- Long labels should wrap rather than get truncated. They don't exist in the data currently, but you can test it by editing one of them. There are [examples of this in the Figma design](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15356), e.g. "Median Household Annual Income" wraps to two lines and slightly increases the height of its row.

- TableView should get a scrollbar if its content overflows vertically or horizontally. When the window is short, TableView should not push the branding bar off the bottom of the screen, and when the window is narrow, scrolling horizontally should not move the sidebar.

- I could use some help understanding react-virtualized. If each section of the table is its own react-virtualized Table, does that mean we aren't really taking advantage of react-virtualized? My assumption is that it skips rendering elements that overflow outside of a Table element, but it seems like nothing ever overflows in this setup. If we scale to 100 sections, would 100 Table components always be rendered? And even if Collapsible sections are collapsed, the Table inside isn't aware of that and still gets rendered to the DOM? Does react-virtualized have its own way of doing collapsible sections / rows which we could use instead of react-collapsible's Collapsible in order to use one big virtualized Table instead of many small ones?

- It looks like rowHeight is static. Does react-virtualized support dynamic row heights? Some will be taller than others given their visualizations, and others will vary based on the length of their label.

src/index.js
```diff
@@ -3,6 +3,7 @@ import ReactDOM from "react-dom";
 import App from "./App";
 import { Provider } from "react-redux";
 import { store } from "./redux/store";
+import 'react-virtualized/styles.css';
```
justinanderson, line 6: This is now an invalid import and leads to a runtime crash.

src/css/TableView.css
```diff
@@ -0,0 +1,60 @@
+.table-view {
+  overflow: scroll;
+  margin-left: auto;
+  margin-right: auto;
+  margin-top: 50px;
```
justinanderson, line 5: This should be `padding-top: 50px` instead of `margin-top`. We want the scrollable area to include the area underneath `Header`.

justinanderson (changes requested):
In addition to line comments:

1. If the table columns are going to squish as the screen width shrinks, they should squish by equal amounts. Columns should stay aligned. Right now, if I make my window narrow, the location names at the top are no longer aligned with their details below. The same goes for the contents of Demographic Summary, it squishes more than the other sections beneath it. I'd be inclined to fix the width of columns and not let them shrink or expand for this first release of the project.

2. Location names at the top of the table should be bottom-aligned instead of top-aligned, per the [Figma design](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15356).

3. There should be more vertical space between location names at the top of the table and the collapsible sections beneath them.

4. When someone removes the last location from "Locations to Compare", let's have that work the same as clicking the back arrow next to "Comparing Locations" and leave comparison mode. Right now, clearing all of the compared locations leave the user in a weird state.

5. TableView's propType for `comparisonFeatures` is wrong. It's object but should be array. I see a runtime warning about this.

justinanderson (changes requested):
- `.table-header`'s font-size should be 16px, not 14.
- `.table-header`'s font-weight should be medium, not bold, so `font-weight: 500`.
- table labels and cells should be 16px, not 15.
- table labels and cell contents should be vertically centers, so `.table-row > td` should get `vertical-align: middle;`
- Also, some requested changes from the last review haven't been made yet:
    - Uneven squishing of table columns. This should probably be resolved by setting a fixed width for columns, since each Collapsible has its own table element and getting those to scale their columns equally would be a headache. 190px for the label column and 140px for data columns would match the Figma design.
    - There needs to be more vertical space between the location names in the table's header and the Collapsible beneath it. `.table-header` should get `margin-bottom: 5px`.

justinanderson (approved):

PR #123, opened 2021-05-25
header left edge matches legend

Tibblist:


src/components/header.js
```diff
@@ -28,10 +28,11 @@ function Header () {
     justifyContent: "space-between",
     alignContent: "center",
     alignItems: "center",
+    right: 0
   };
 
   return (
-    <div className="container" style={headerBox}>
+    <div className="container sidebarOpen" style={headerBox}>
```
justinanderson, line 35: This needs to be conditional, or else the sidebar is always considered to be open. Also, make sure it works in comparison mode as well.

And it's probably better to move [that logic](https://github.com/CxSci/SA-dashboard/blob/master/src/pages/main.js#L26-L29) from main.js to app.js, the parent it shares with header.js, instead of duplicating logic between sibling components.

justinanderson (changes requested):

src/components/header.js
Tibblist: Works in comparison mode even though the header appears to be empty in comparison mode.

PR #130, opened 2021-06-01
Grid view

### LINKED ISSUE #125: Add a grid view to comparison mode
Note this task depends on the Table / Grid control from #124.

In addition to the comparison mode's default table view, there should be an alternative grid view. Ideally, it would reuse the component used to display multiple locations' detail in the sidebar when `comparisonFeatures` is populated. [Figma mockup](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15440).

gerardott:
Closes #125


src/components/GridView.js
```diff
@@ -0,0 +1,52 @@
...
+import { connect } from "react-redux";
+import Collapsible from "react-collapsible";
+import propsMapping from "./propsMapping";
+import MetricDetails from "./MetricDetails";
+import "../css/GridView.css"
+
+const GridView = ({comparisonFeatures}) => {
+  const comparisonFts = comparisonFeatures;
+
+  const renderTable = (section) => {
```
justinanderson, line 12: Probably should be called `renderGrid` instead of `renderTable`.

src/components/GridView.js
```diff
@@ -0,0 +1,52 @@
...
+    return (
+      <div className="grid-container">
+        {section.content.map((metric) => (
+          <div key={metric.id} className="grid-item">
+            <div className="grid-item-head">
+              <h2>{metric.label}</h2>
+            </div>
+            <div className="grid-item-body">
+              {comparisonFts.map(ft => 
+                <MetricDetails key={ft.id} feature={ft} metric={metric} />
```
justinanderson, line 22: Same comment as in LocationDetails.js and MetricDetails.js about using `ft.SA2_MAIN16` instead of `ft.id`.

justinanderson (changes requested):
1. The grid view's contents should be allowed to wrap. Removing the fixed width from `.grid-view` should accomplish that.

2. Font size of metric labels should be 14px instead of 15px and weight should be 700 instead of 800. The same is true for when they're in the sidebar. That's defined in collapsible.css using the h2 selector, and that should probably be changed to a more specific selector like `.Collapsible__contentInner h2`.

3. Per the Figma design, let's limit region names in grid-items to a single line. That can be accomplished with something like `.grid-item-body .comparison span:first-child { text-overflow: ellipsis; white-space: nowrap; overflow: auto; margin-right: 8px; }`. 

4. There should be a little more whitespace above `.grid-container .grid-item` — a 5px margin-top would do it.

5. Less bottom margin on `.grid-container .grid-item-body`. 10px instead of 30. It ends up being too much when a grid wraps to 2 rows. 

6. Change `.grid-container .grid-item-head` and `.grid-container .grid-item-body` horizontal padding from 20px to 15px.

7. Metrics in Figma were designed to be 250px wide, not including the extra padding around separators, so the width of `.grid-container .grid-item` should be increased to 260px to compensate for that extra 5px of horizontal padding on each side.

8. A max-width of 1040px for the grid view would be good, to make sure it's never more than 4 grid-items across on very large windows.

9. There should be some horizontal padding on the overall grid view so it never overlaps the sidebar toggle. Also, there should be slightly more vertical padding at grid view's top. Changing `.grid-view`'s `padding-top: 60px;` to `padding: 65px 25px 0;` should do it.

Also, the separators between grid-items should be sized and positioned differently, and there should be horizontal lines separating grid rows when they wrap, but what's here now is good enough for a first pass. That can be dealt with in another PR in the future if we have time.

justinanderson (approved):

Issue #126, opened 2021-05-26
Add compared features to URL


justinanderson:
We should add locations being compared to the URL, so that reloading the comparison page doesn't result in an error. That'd make it easier to finish implementing comparison mode as well. A space-delimited list of each feature's SA2_MAIN16 would be nice, e.g. `/comparison/401021003+405011110+407031164+402031037`.

PR #131, opened 2021-06-03
Recommendation tool

Tibblist:
This isn't quite done, just need to make state save input for each question and possibly style the inputs


src/components/RecommendationTool.js
```diff
@@ -0,0 +1,113 @@
...
+}
+
+const buttonText = {
+    "fontFamily": "Roboto",
+    "fontStyle": "normal",
+    "fontWeight": "bold",
+    "fontSize": "16px",
+    "lineHeight": "19px",
+    "color": "#FFFFFF"
+}
```
justinanderson, line 63: This and the "Recommendation Tool" button in RecommendationModal.js are the sort of situations I had in mind for a reusable ComparisonButton. It'd need a prop for turning off the chevron in the "Previous" and "Next" cases and a prop for styling it blue instead of an outline, but then there wouldn't be any need to hardcode widths and heights. Also, it'd make it easier to do an accessibility pass in the future if there are fewer button implementations to review.

src/components/RecommendationTool.js
```diff
@@ -0,0 +1,113 @@
...
+    const [formState, setFormState] = useState({});
+    return <>
+        <RecommendationHeader currentStage={currentStage} stages={props.data.map(x => x.title)}/>
+        <div style={root}>
+            <p style={descriptionText}>{props.data[currentStage].description}</p>
+            {props.data[currentStage].questions.map(question => {
+                let inputComponent = <></>
+                switch (question.type) {
+                    case "multiple_choice":
+                        inputComponent = <>{question.answers.map(answer => <><label key={answer} style={labelRoot}><input type="radio"/>{answer}</label><br/></>)}</>;
```
justinanderson, line 82: Radio buttons need the same `name` attribute to act like a radio group. Looking at constant.js, that'd be `name={question.key}` here.

justinanderson, line 82: I get a `Warning: Each child in a list should have a unique "key" prop.` on this line.

src/constants.js
```diff
@@ -22,3 +22,70 @@ export const MAP_TYPE = {
...
+      {
+      type: "checkbox",
+      question: "What characteristics describe your typical customers?",
+      hint: "Select all that apply",
+      key: "characteristics",
+      answers: ["Tourists", "Locals", "One-off Purchases", "Repeat Customers", "Culturally Diverse"]
+    },
+    {
+      type: "checkbox",
+      question: "Waht's their typical household income?",
```
justinanderson, line 72: Waht

justinanderson (changes requested):
Various comments, understanding this is still in progress. Also, it'll be good if the dummy version of this includes a dummy result, like the last [Figma step](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1561%3A493), which could eventually be hooked up to go straight into comparison mode once #126 is finished.

src/pages/recommendation.js
```diff
@@ -0,0 +1,48 @@
...
+
+const Recommendation = () => {
+    const history = useHistory()
+    return <div style={root}>
+        <div style={titleRow}>
+        <p style={titleText}>Recommendation Tool</p>
+            <CloseIcon style={iconButton} onClick={()=>{ history.goBack() }}/>
+        </div>
+        <RecommendationTool data={RECOMMENDATION_DUMMY_DATA}/>
+    </div>
```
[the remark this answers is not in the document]
Tibblist: Do you want me to change container.js to use .goBack() or change the tool to use container.js instead of how it is currently setup?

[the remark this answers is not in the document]
justinanderson: Both

PR #135, opened 2021-06-11
Data visualization

### LINKED ISSUE #133: Render data visualizations for some metrics
It's time to render metrics as more than just text.

We're still working out what data we'll be able to present and what visualizations will be appropriate for those, but at a minimum we'll need:

- [Bar charts](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1479%3A11239)
- Ranges, both [single](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1479%3A11140) and [multiple](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1479%3A12968)

For now, it's a fair guess that normalized properties like `gini_coefficient_no` should be displayed as a range, and bar charts should be used when comparing quantities like `persons_num`.

gerardott:
Closes #133


src/components/charts/RangeBar.js
```diff
@@ -0,0 +1,100 @@
+import React from "react";
```
justinanderson, line 1: RangeBars should handle `NaN` better, e.g. "Le Hunte - Elliston":

Screen Shot 2021-06-21 at 12 16 51 PM

Ideally they'd have some way for the range to show the value is undefined, but for now let's just replace the RangeBar with the text "Undefined" when a value is `NaN`.

src/components/charts/RangeBar.js
```diff
@@ -0,0 +1,100 @@
...
+      <Scatter dataKey="value"
+        shape={renderDashedLine}
+        isAnimationActive={false}
+      />
+      <Scatter dataKey="value"
+        shape={renderTriangleShape}
+        isAnimationActive={false}
+      />
+      <ReferenceArea x1="0" x2={data.max} fill="transparent">
+        <Label value="LOW" position="insideLeft" fontSize={10} fontWeight={500} />
```
justinanderson, line 49: The "LOW" and "HIGH" strings should come from propsMapping, just like `min` and `max`. Same with the gradient's stop colors. Those may vary from metric to metric.

That said, it'd be good to have "LOW" and "HIGH" and the current colors as defaults in case those aren't defined in propsMapping.js.

src/components/charts/LineChartMetric.js
```diff
@@ -0,0 +1,72 @@
...
+    dataKey: `value`,
+    data: item.data,
+    name: truncate(item.name, 20),
+    color: getColorFromGradient('rgb(255,233,36)', 'rgb(242,11,11)', (idx + 1) / 5)
+  }));
+
+  if (showLegend) height += (lines.length * 15);
+
+  return (
+    <LineChart width={width} height={height}
```
justinanderson, line 33: I can only visit the comparison mode by visiting a pre-built comparison link directly. If I do the normal thing of visiting `/`, adding feature to compare, and then clicking "Show Comparison", I get a blank screen and console errors like `TypeError: undefined is not a function (near '...displayedData.map...')`.

src/components/MetricDetails.js
```diff
@@ -1,23 +1,63 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { formatValue } from "../utils/formatValue";
+import RangeBar from "./charts/RangeBar";
+import SolidBar from "./charts/SolidBar";
+import LineChartMetric from "./charts/LineChartMetric";
+
+const MetricDetails = ({ featureList, metric, small }) => {
+  const renderSingleMetic = (feature) => {
```
justinanderson, line 9: Missing an 'r'

src/components/charts/LineChartMetric.js
```diff
@@ -0,0 +1,72 @@
...
+  XAxis,
+  YAxis,
+  CartesianGrid,
+  Legend,
+  Tooltip,
+} from "recharts";
+import { getColorFromGradient } from '../../utils/colors';
+const truncate = (str, n) => (str.length > n) ? str.substr(0, n-1) + '…' : str;
+
+const LineChartMetric = ({ width = 260, height = 140, series, data, showLegend }) => {
```
justinanderson, line 15: Let's try `height = 180` as a default. 140 feels too cramped.

src/components/charts/LineChartMetric.js
```diff
@@ -0,0 +1,72 @@
...
+  
+  const lines = series.map((item, idx) => ({
+    key: item.name,
+    dataKey: `value`,
+    data: item.data,
+    name: truncate(item.name, 20),
+    color: getColorFromGradient('rgb(255,233,36)', 'rgb(242,11,11)', (idx + 1) / 5)
+  }));
+
+  if (showLegend) height += (lines.length * 15);
```
justinanderson, line 30: This doesn't seem to have the desired effect:
Screen Shot 2021-06-21 at 12 47 07 PM

The chart seems to take up a fixed percentage of the total height, instead of using what's left after the legend is laid out.

This might be solvable using Legend's `wrapperStyle` and some careful CSS.

justinanderson (changes requested):

src/components/LocationDetails.js
```diff
@@ -30,15 +30,31 @@ const LocationDetails = (props) => {
...
     )
   }
+  
+  const renderSelectedValue = (metric) => {
+    let rawValue = selectedFeature.properties[metric.id];
+    const value = formatValue(rawValue, metric.format);
+    
+    switch (metric.type) {
+      case 'line-chart': {
+        const data = JSON.parse(rawValue);
```
justinanderson, line 47: This crashes with `SyntaxError: JSON Parse error: Unexpected identifier "object"` if I search for anything when there's nothing in `comparisonFeatures`.

src/components/charts/LineChartMetric.js
justinanderson: Some further information: going to comparison and back again makes this render a little more correctly. So this might be an issue with the chart not redoing its layout when its data changes.

justinanderson: Other people having a similar issue with some suggested workarounds https://github.com/recharts/recharts/issues/511.

justinanderson (approved):

PR #136, opened 2021-06-14
Business layer

Tibblist:


src/components/legend.js
```diff
@@ -21,14 +21,21 @@ const mapTypes = [
     value: Constants.MAP_TYPE.TRANSACTIONS,
     text: "Financial Interactions - Growth Potential",
   },
+  {
+    value: Constants.MAP_TYPE.BUSINESS,
+    text: "Business"
```
justinanderson, line 26: "Businesses"

justinanderson (changes requested):
I'm seeing clusters when zoomed out, but then nothing when zoomed in close enough to not need clustering. Half of the issue remains unimplemented.

Tibblist:
Somehow I had missed the other reqs, almost all should be added. Working on popup effects soon and should have that done tonight.

justinanderson (changes requested):
In addition to line comments:

- Make the minimap use the economic layer if the businesses layer is active, since it doesn't draw correctly in comparisons otherwise.

- Make the map still draw businesses after leaving comparison mode. Right now, `drawBusinessClusters()` is only called in `componentDidUpdate()`, so if you enter comparison mode while viewing the "Businesses" layer and then come back, the new map isn't aware that its active layer isn't one of the map-fill ones. There's actually a console error coming back indicating that: `Error: layers.sa2-fills.paint.fill-color.stops: array must have at least one stop`.

- Clear popups when switching layers. If I click on a business and then switch to another layer, the popup for that business still shows.

src/components/map.js
```diff
@@ -211,29 +227,91 @@ let Map = class Map extends React.Component {
...
+    const properties = e.features[0].properties;
+
+    // Ensure that if the map is zoomed out such that
+    // multiple copies of the feature are visible, the
+    // popup appears over the copy being pointed to.
+    while (Math.abs(e.lngLat.lng - coordinates[0]) > 180) {
+      coordinates[0] += e.lngLat.lng > coordinates[0] ? 360 : -360;
+    }
+
+    new mapboxgl.Popup()
```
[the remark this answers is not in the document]
Tibblist: Have this mostly done, do we need the little arrow portion of the popup bubble. Not really sure how to accomplish that I know there are some css tricks that might do it using :before/:after and I don't see an svg for it

src/components/map.js
```diff
@@ -2,7 +2,10 @@ import React from "react";
 import PropTypes from "prop-types";
```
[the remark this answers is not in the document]
Tibblist: Got it, was getting messy and figured I'd hit the file with it since I wasn't aware of a way to do it only for certain sections but I'll try to find something that is more surgical

src/components/map.js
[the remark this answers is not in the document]
justinanderson: That's what `.mapboxgl-popup-tip` is. If you do the CSS stuff I recommended, it should be pretty simple.

Tibblist:
A bug I am noticing is selecting a feature when in business mode crashes the app, presumably this is because the point features don't have all the same properties. Not sure if this bug will apply in the real app because points will be made to have the same vs the sample data or if this needs to be addressed somehow.

justinanderson:
Ah yeah, that's something else to worry about. When you're on the "Businesses" layer, you can't click on map to select features but you can still select them via the search field. Maybe have map.js's `selectFeature()` for the map into one of the other layers if you're on the business layer? I'm assuming the crash happens there.

Tibblist:
For me it actually crashes when selected via search, was that not happening for you?

justinanderson:
> For me it actually crashes when selected via search, was that not happening for you?

That's what I mean. When you're on the "Businesses" layer, there's no possible way to fire off `selectFeature()` by clicking on the map, but you can still do so via the search field, leading to the crash.

Tibblist:
Also for showing econ layer when in comparison mode, is it fine to just switch the legend state to econ? This would let me pull in the same stops and not have to hardcode them into map, otherwise I would need to hardcode the econ fill information in map.js to only show in mini/comparison mode.

justinanderson:
Sorry, I meant to write the Mobility layer, not Economic. And yeah, that'll be fine.

src/components/map.js
```diff
@@ -281,108 +361,239 @@ let Map = class Map extends React.Component {
...
     // Skip if the selection hasn't actually changed or if the map isn't fully loaded yet
-    if (prevId === newId || !this.map.getSource("sa2") || !this.map.isSourceLoaded("sa2")) {
-      return
+    if (
+      prevId === newId ||
+      !this.map.getSource("sa2") ||
+      !this.map.isSourceLoaded("sa2")
+    ) {
+      return;
+    } else if (this.props.active.key !== Constants.MAP_TYPE.BUSINESSES) {
```
justinanderson, line 577: This prevents the selection from being drawn, but it doesn't prevent the highlight from lingering with no way of being dismissed. The mouse handler that normally does that is part of the `sa2-fills` map layer. Rather than deal with any of that right now, let's say that, for now, if someone manages to call `selectFeature()` while on the "Businesses" layer, we'll make the map/legend switch to the "Mobility" layer to force them to use the code that currently works without odd usability issues.

src/components/map.js
```diff
@@ -281,108 +361,239 @@ let Map = class Map extends React.Component {
...
+
+    if (!this.map.getLayer("cluster-count"))
+      this.map.addLayer({
+        id: "cluster-count",
+        type: "symbol",
+        source: "business",
+        filter: ["has", "point_count"],
+        layout: {
+          "text-field": "{point_count_abbreviated}",
+          "text-font": ["DIN Offc Pro Medium", "Arial Unicode MS Bold"],
```
justinanderson, line 456: It's worth adding a comment here for posterity that Mapbox uses its own fonts and doesn't support this project's default Roboto.

justinanderson (changes requested):

src/css/map.css
```diff
@@ -38,3 +39,25 @@
...
+  font-family: Roboto;
+  font-style: normal;
+  font-weight: normal;
+  font-size: 14px;
+  line-height: 250%;
+  color: #000000;
+}
+
+.popup-root {
+  padding: 15px;
```
justinanderson, line 62: I'm guessing you copied 15px padding from the Figma design, but there's also padding on `.mapboxgl-popup-content` which ends up making this too much padding. It'd be better to set `padding: 15px;` on `.mapboxgl-popup-content` using a more specific selector after you give the popup its own className.

src/components/map.js
```diff
@@ -211,29 +227,96 @@ let Map = class Map extends React.Component {
...
+    // Ensure that if the map is zoomed out such that
+    // multiple copies of the feature are visible, the
+    // popup appears over the copy being pointed to.
+    while (Math.abs(e.lngLat.lng - coordinates[0]) > 180) {
+      coordinates[0] += e.lngLat.lng > coordinates[0] ? 360 : -360;
+    }
+
+    new mapboxgl.Popup()
+      .setLngLat(coordinates)
+      .setHTML("<div class=\"popup-root\"><p class=\"popup-text-header\"> " + properties.name + "</p><br><p class=\"popup-text\">Type: " + properties.type + "</p></div>")
```
justinanderson, line 286: If you get rid of the `<br>` here, there would be about the right amount of vertical space between the name and type.

Also, it'd be good to not include the text "Type: " as part of this.

justinanderson (changes requested):

src/components/map.js
Tibblist: made it switch to mobility layer if select feature is called on business layer but it doesn't seem to update the legend. This might be due to store dispatching manually inside the file vs this.props.dispatch but I'm not sure yet

justinanderson: It turns out that's due to a bug in DropdownSelect. By using `initialSelectedItem` instead of `selectedItem` on that, it's ignoring changes to the `activeLayer` state. I'll push a fix for that shortly.

PR #139, opened 2021-06-23
Add slider input

Tibblist:
In progress PR for rc-slider issue


Tibblist:
By the way I kept the answers input as an array that got turned into an object {0: "", 20: "" etc} to make it simpler for someone editing the json to just put an array of possible answers as with others and the component handles transforming it to marks but this will work fine too.

justinanderson:
> By the way I kept the answers input as an array that got turned into an object {0: "", 20: "" etc} to make it simpler for someone editing the json to just put an array of possible answers as with others and the component handles transforming it to marks but this will work fine too.

Yeah, I went back and forth on that. It would be nice to not mix the 0 - 100 implementation detail of rc-slider in with the questions themselves, but at the same time it's useful to have the freedom to put slider marks anywhere on the slider, not just evenly spaced out.

Issue #147, opened 2021-06-25
Render textual metrics how they're laid out in Figma


justinanderson:
Render textual metrics like in the [Figma design](https://www.figma.com/file/vP3gxGjxW34i5gNMpu6sDW/Design-V1?node-id=1517%3A15279). Note how the values for "Population" and "Median Household Income" are right aligned and on the same line as their labels.

Here's a screenshot of how it master looks right now versus how it should look:
<img width="273" alt="Screen Shot 2021-06-24 at 8 52 04 PM" src="https://user-images.githubusercontent.com/244952/123353036-9cc06b00-d52e-11eb-89fe-2ef7ff16bbd3.png"><img width="315" alt="Screen Shot 2021-06-24 at 8 55 20 PM" src="https://user-images.githubusercontent.com/244952/123353038-9d590180-d52e-11eb-8f7b-45487c9f15df.png">

justinanderson:
Resolved by #152.

PR #152, opened 2021-06-28
No data in linechart

### LINKED ISSUE #141: Do something in LineChart when there's no data to display
We don't have "Projected Population" data for some regions (e.g. "The Parks", "Happy Valley", "West Coast (SA)"), so sometimes LineChart is blank.

It's probably ok if a chart has one visible line and the rest are missing, but if a chart ends up with no data to display, maybe superimpose the text "No data" in the center of the chart?

So instead of this:
Screen Shot 2021-06-24 at 5 25 40 PM

We would have this:
Screen Shot 2021-06-24 at 5 25 40 PM

gerardott:
Closes #141 and #147

### DIFF
--- a/src/components/LocationDetails.js
+++ b/src/components/LocationDetails.js
@@ -40,7 +40,7 @@ const LocationDetails = (props) => {
   
   const renderSelectedValue = (metric) => {
     let rawValue = selectedFeature.properties[metric.id];
-    if (rawValue === undefined || rawValue === null || (Array.isArray(rawValue) && rawValue.length === 0)) {
+    if (rawValue === undefined || rawValue === null) {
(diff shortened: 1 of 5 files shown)


justinanderson (changes requested):
Could you make it so single locations without data get empty line charts instead of the text "n/a" or "No data"?

So, instead of selecting "West Coast (SA)" and seeing:
Screen Shot 2021-06-28 at 12 12 25 PM

See this instead:


And the same for table view comparisons. It'd be preferable for the "West Coast (SA)"'s projected population to use a "No data" line chart instead of just the text:
Screen Shot 2021-06-28 at 12 13 34 PM

justinanderson (approved):

PR #194, opened 2022-02-16
SA-192 small business support visualizations

### LINKED ISSUE #192: Create Small Business Support data visualizations
Create the data visualizations which will be used by the Small Business Support dataset.

The layout of Atlas v2's detail pages will be dynamic rather than hardcoded. That means two separate API responses define what is rendered on the detail page, the raw data of the entry and the layout description which describes how to present that data to the user.

There is an example of each in the [frontend/mocked_api_responses](https://github.com/CxSci/SA-dashboard/tree/master/frontend/mocked_api_responses) directory: [data for Adelaide](https://github.com/CxSci/SA-dashboard/blob/master/frontend/mocked_api_responses/detail_data_example_small_business_support_adelaide.json) and a [layout description for Small Business Support](https://github.com/CxSci/SA-dashboard/blob/master/frontend/mocked_api_responses/detail_layout_example_small_business_support.json). There will eventually be an API endpoint on the backend which returns JSON responses like those. These examples represent what would be rendered when a user visits /explore/small-business-support/401011001.

There's also [some documentation](https://github.com/CxSci/SA-dashboard/blob/master/frontend/mocked_api_responses/detail_layout_description.md) about the format of the layout description file, for reference.

There's a [Figma mockup](https://www.figma.com/file/yaKUmuZM9xPQ1Qq5YVf5Yu/?node-id=2892%3A29536) of what detail_layout_example_small_business_support.json should look like as well. The goal of this issue isn't to completely finish the detail page, just to replicate what's at that Figma link using the provided data.

Feel free to move and/or rename the mocked_api_responses folder to wherever you think makes most sense.

leo-esaki:
Related to #192


frontend/src/routes/SmallBusinessSupport/SmallBusinessSupport.js
```diff
@@ -0,0 +1,41 @@
...
+
+const SmallBusinessSupport = () => {
+  const { sections } = sectionsLayout
+  return (
+    <Container sx={{ mt: 3 }}>
+      {sections.map((section, index) => (
+        <Box key={index} sx={{ mb: 2 }}>
+          <Typography variant="h4" gutterBottom>
+            <strong>{section.title}</strong>
+          </Typography>
```
justinanderson, line 21: I think I'd prefer adding custom variants to the MUI theme when possible rather than specifying much CSS in components. For example, this could be `<Typography variant="sectionTitle">{section.title}</Typography>`. Similarly, `text`, `number`, and `simple_bar` are all meant to share text styles in the Figma design.

I'm almost definitely going to go back and forth tweaking fonts and spacing of everything a few times on this project, and having much of the project's appearance controlled by theme.js would make that easier for me.

Also, it's likely that some deployers of Atlas v2 will want to style/theme their instance to differentiate it from our main one, and I'd like to make that straightforward for them if possible.

frontend/.storybook/main.js
```diff
@@ -4,5 +4,7 @@ module.exports = {
   framework: '@storybook/react',
   core: {
     builder: 'webpack5'
-  }
+  },
+  managerWebpack: updateEmotionAliases,
+  webpackFinal: updateEmotionAliases
```
justinanderson, line 9: Storybook dies here with `ReferenceError: updateEmotionAliases is not defined`.

justinanderson:
Rebased onto master (force push) and merging this PR so shareable work and dependencies can be shared with other developers.

Issue #193, opened 2022-02-15
Add a navigation header


justinanderson:
Every page of the Atlas has a navigation header fixed at the top. The header's contents can vary depending on the page and whether the page has been scrolled vertically, but there's enough similarity between the different versions that they should all share a base component.

The navigation header should support two styles, one light and one dark. In the designs so far, only the edge-to-edge map will use the dark style.

Some useful Figma links for this issue:

- [Examples of navigation headers in use](https://www.figma.com/file/yaKUmuZM9xPQ1Qq5YVf5Yu/?node-id=2502%3A25123)
- [Standalone examples of navigation headers](https://www.figma.com/file/yaKUmuZM9xPQ1Qq5YVf5Yu/Design---Skills?node-id=2908%3A34793)
- [Standalone examples of header breadcrumbs](https://www.figma.com/file/yaKUmuZM9xPQ1Qq5YVf5Yu/Design---Skills?node-id=2914%3A35481)
- [Standalone examples of search fields](https://www.figma.com/file/yaKUmuZM9xPQ1Qq5YVf5Yu/Design---Skills?node-id=2914%3A36065)

Note that the Figma designs aren't completely accurate. Figma doesn't support putting borders or shadows on only one side of a shape, but try to imagine that navigation headers with shadows only have a shadow on their bottom edge.

Also, don't worry about implementing all of the contents of the "Scrolled Down" headers from the standalone examples. For this issue, it'll be enough if the header has working breadcrumbs and supports switching its appearance when pageYOffset > 0, regardless of what that other appearance looks like.

torosyaneric:
Remaining to implement: **expanded view with content**

torosyaneric:
PR: https://github.com/CxSci/SA-dashboard/pull/197

PR #195, opened 2022-02-17
Add Material-UI + Header component #193

torosyaneric:


justinanderson:
Rebasing into master so this basic navigation UI and material-ui are available for other developers.

PR #197, opened 2022-02-21
Header changes

torosyaneric:
[Change header search to autocomplete](https://github.com/CxSci/SA-dashboard/commit/cdc72f7a5a1d8381ed14b168c905bdf52cbed65a)
[Update compare buttons text](https://github.com/CxSci/SA-dashboard/commit/3d72aaec4856710f5c2fe8f0170eb9aa024b0d26)


justinanderson (changes requested):
Ok, now that the basics are in place I'll be more nitpicky.

- [x] Please disable MUI's ripple effect, https://mui.com/getting-started/faq/#how-can-i-disable-the-ripple-effect-globally.

- [x] On browsers other than Chrome, `pageScrolled` is incorrect if you reload a page which is already scrolled down. It looks like Chrome is the only browser which fires a `scroll` event after a reload. You can fix this if you add the following to Header.container.js:
    ```js
    // Call scroll callback on page load to ensure `pageScrolled` is accurate.
    // Firefox and Safari report `window.scrollY` as 0 on page load and don't fire
    // an initial scroll event even if the page is already scrolled due to a reload
    // or a URL hash.
    useEffect(() => {
      onScroll()
    }, [onScroll])
    ```

- [x] Pages should use the light theme by default and be able to override that . So far, only /explore/small-business needs to use the dark theme. This will be defined by the backend as part of the list of available datasets. For now, I recommend adding a `theme: 'dark'` to `'small-business'` in `DATASETS_MAP`.

- [x] Similarly, header's `scrolledHeight` should default to being the same as `height`, and pages should be able to override that. It should default to staying 60px. For instance, header height shouldn't change on the home page or static pages like the FAQ.

- [x] Header should be slightly translucent: `backgroundColor: rgba(255, 255, 255, 0.9)` in light mode and `backgroundColor: rgba(0, 0, 0, 0.3)` in dark mode.

- [ ] The home page's header is a special case. On the home page, AppContainer's `paddingTop` should be 0, and until the page is scrolled the header's background should be transparent and without a drop shadow. In fact, it might make sense to delete AppContainer, give Dashboard its own wrapping div, and move AppContainer's paddingTop to that. Here are examples of how the homepage should look, [unscrolled](https://user-images.githubusercontent.com/244952/155022031-5b6363ab-5d8e-4df7-b296-dd9eb6596af9.png) and [scrolled](https://user-images.githubusercontent.com/244952/155022311-f99e3d3f-1d87-4a32-97ff-a5151b69aa43.png).

- [x] Breadcrumbs should be 24px (`1.5rem`), not 16px, and in light mode they should be `#333333`. Also, their separators should be styled the same way.

- [x] The menu/arrow icons:
    - Icons should be `fontSize="large"`.
    - `<IconButton size="large" sx={{ mr: 1 }}>` would be closer to the Figma design.
    - In light mode, they should be `#4d4d4d` normally and `#000000` when hovered.
    - Shouldn't have a circular background highlight when hovered.

- [x] The "Add to Comparison" and "Compare" buttons:
    - The buttons' titles should have `fontWeight: 700`.
    - On hover, the buttons should have `backgroundColor: '#36ABE5'` and no drop shadow.
    - The buttons shouldn't resize or wrap their titles when the page gets narrow or the breadcrumbs get long.
    - Similarly, the buttons' icons shouldn't shrink. Narrowing the window on /explore/small-business/dataset is a good example of this.

frontend/src/utils/theme.js
```diff
@@ -152,12 +179,13 @@ const initTheme = darkMode =>
...
       header: {
-        height: '60px',
+        height: headerHeight,
         scrolledHeight: '84px',
         paddingX: '14px',
         paddingY: '10px',
         boxShadow: '0px 2px 4px rgba(0, 0, 0, 0.25)',
-        bgColor: darkMode ? 'rgba(0, 0, 0, 0.1)' : '#fff',
+        bgColor: darkMode ? 'rgba(0, 0, 0, 0.3)' : 'rgba(255, 255, 255, 0.9)',
+        iconColor: darkMode ? '#4d4d4d' : '#000000',
```
justinanderson, line 188: Menu and arrow icon buttons should be `'#ffffff'` in dark mode and `'#4d4d4d'` in light mode. And when hovered, they should be `'#ffffff'` in dark mode and `'#000000'` in light mode.

frontend/src/utils/theme.js
```diff
@@ -152,12 +179,13 @@ const initTheme = darkMode =>
       },
 
       header: {
-        height: '60px',
+        height: headerHeight,
         scrolledHeight: '84px',
```
justinanderson, line 183: Would it make sense to change this to `scrolledHeight: scrolledHeaderHeight`?

justinanderson (changes requested):
- [ ] Header background should be [completely transparent](https://user-images.githubusercontent.com/244952/155022031-5b6363ab-5d8e-4df7-b296-dd9eb6596af9.png) on home page when not scrolled (`backgroundColor: 'transparent'`).

- [x] The Research page is missing a header.

- [x] Static pages like FAQ, Research, and Contributors should have breadcrumbs. For instance, the FAQ page's breadcrumbs would be `Atlas of Opportunity / FAQ`. I've updated the Figma design with an [example](https://www.figma.com/file/yaKUmuZM9xPQ1Qq5YVf5Yu/Design---Skills?node-id=3153%3A29944).

justinanderson (approved):

PR #198, opened 2022-02-22
SA-192 Small business support visualizations (part 2)

leo-esaki:
Related to #192

Implemented SimpleRange section

### DIFF
--- a/frontend/src/components/SimpleRange/SimpleRange.js
+++ b/frontend/src/components/SimpleRange/SimpleRange.js
@@ -12,7 +12,7 @@ const GradientProgress = styled(LinearProgress)(({ theme, value }) => ({
       display: 'block',
       width: '100%',
       height: '100%',
-      borderRadius: 1,
+      borderRadius: theme.shape.borderRadius,
       transition: 'transform .4s linear',
       transform: `translateX(${100 - value}%)`,
       background: `linear-gradient(270deg, ${theme.palette.canary.main} 0%, ${theme.palette.chestnutRose.main} 50%, ${theme.palette.ultramarine.main} 100%)`, // eslint-disable-line
@@ -26,7 +26,12 @@ const SimpleRange = ({ min, max, value, style, color }) => {
   return (
     <Range
       value={percentage * 100}
-      sx={{ height: 24, borderRadius: 1, [`& .${linearProgressClasses.bar}`]: { borderRadius: 1 } }}
+      sx={{
+        height: 24,
+        borderRadius: 1,
+        [`& .${linearProgressClasses.bar}`]: { borderRadius: 1 },
+        '-webkit-mask-image': '-webkit-radial-gradient(white, black)', // fixes border-radius on safari
+      }}
       variant="determinate"
       color={color}
     />


frontend/src/components/SimpleRange/SimpleRange.js
```diff
@@ -0,0 +1,48 @@
...
+      height: '100%',
+      transition: 'transform .4s linear',
+      transform: `translateX(${100 - value}%)`,
+      background: `linear-gradient(270deg, ${theme.palette.canary.main} 0%, ${theme.palette.chestnutRose.main} 50%, ${theme.palette.ultramarine.main} 100%)`, // eslint-disable-line
+    },
+  },
+}))
+
+const SimpleRange = ({ min, max, value, style, color }) => {
+  const Range = style === 'gradient' ? GradientProgress : LinearProgress
```
justinanderson, line 24: Both the outer and inner bars of SimpleRanges should have rounded corners. The inner bar is missing them in all browsers and there's a rendering difference in Safari which makes the outer bar not have rounded corners.

You can fix all of that if you add `borderRadius: '4px'` to LinearProgress-bar, `borderRadius: '4px'` to GradientProgress's `'&:before'`, and use GradientProgress for both gradient and solid style SimpleRanges.

That last bit is required to make the left-side corners rounded on Safari. Something about putting the fill in a `'&:before'` pseudo-element makes Safari round its corners properly.

frontend/src/components/SimpleRange/SimpleRange.js
```diff
@@ -0,0 +1,48 @@
...
+  const percentage = Math.min(1, (value - min) / (max - min || 1))
+  return (
+    <Tooltip title={<FormattedNumber value={value} />} placement="top">
```
justinanderson, line 27: We actually don't want tooltips on ranges. Ranges will be for situations where raw values can mislead users. For example, if someone sees GDP Growth Potential is 1.217 they might think that means "121.7% growth", when in truth the number is meaningless by itself. It only makes sense when compared to the whole range.

frontend/src/components/SimpleBar/SimpleBar.js
```diff
@@ -1,6 +1,9 @@
 import Box from '@mui/material/Box'
```
justinanderson, line 1: Similar to the bit about rounded corners in SimpleRange, SimpleBar's corners aren't round in Safari. That can be fixed by applying the same `borderRadius: '4px'` to LinearProgress-bar.

justinanderson (changes requested):

frontend/src/components/SimpleRange/SimpleRange.js
```diff
@@ -1,18 +1,18 @@
...
   [`& .${linearProgressClasses.bar}`]: {
     overflow: 'hidden',
     backgroundColor: 'transparent',
+    borderRadius: 1,
     '&:before': {
       content: '""',
       display: 'block',
       width: '100%',
       height: '100%',
+      borderRadius: 1,
```
justinanderson, line 15: The `borderRadius: 1` on .bar is rendered as `border-radius: 4px`, but the `borderRadius: 1` on ::before is rendered as `border-radius: 1px` instead of the intended 4px. That seems like a bug in MUI to me, but I don't have much experience with it. I would suggest hardcoding the ::before borderRadius as 4px.

frontend/src/components/SimpleRange/SimpleRange.js
```diff
@@ -1,18 +1,18 @@
...
       transition: 'transform .4s linear',
       transform: `translateX(${100 - value}%)`,
       background: `linear-gradient(270deg, ${theme.palette.canary.main} 0%, ${theme.palette.chestnutRose.main} 50%, ${theme.palette.ultramarine.main} 100%)`, // eslint-disable-line
```
justinanderson, line 18: The last bit of from my [previous review](https://github.com/CxSci/SA-dashboard/pull/198#discussion_r812358817) about Safari not rounding all ranges is still outstanding. Adding a borderRadius to GradientProgress fixed it for gradient ranges, but solid ranges still have the problem.

It's the separate ::before element of GradientProgress which fixes the rendering problem on Safari, so the solution that comes to my mind is to use the same solution for solid ranges by making GradientProgress into a function which takes `style` and `color` as inputs and change this line to something like ``background: style === 'solid' ? color : `linear-gradient...`'``.

justinanderson (changes requested):

justinanderson (approved):
Approved, but please avoid using merge commits to keep up to date with master.

We use a rebasing workflow (`git pull --rebase`) for this project's pull requests. Merge commits have a history of creating hard to diagnose bugs and make reviewing pull requests harder. Sometimes merge commit are unavoidable, but those situations should be rare.

PR #201, opened 2022-02-26
SA-192 Implement line chart

leo-esaki:


frontend/src/utils/helpers.js
```diff
@@ -0,0 +1,18 @@
...
+]
+
+export const nFormatter = (num, fractions = 1) => {
+  const rx = /\.0+$|(\.[0-9]*[1-9])0+$/
+  const item = nFormatterLookup
+    .slice()
+    .reverse()
+    .find(item => num >= item.value)
+  return item ? (num / item.value).toFixed(fractions).replace(rx, '$1') + item.symbol : '0'
+}
```
justinanderson, line 18: I think this could be replaced with react-intl's formatNumber. The following returns the same values as nFormatter does, up to 1e12 at least:
```js
import { useIntl } from 'react-intl'
const intl = useIntl()
intl.formatNumber(t, {
  notation: 'compact',
  compactDisplay: 'short'
})
```

It would support localization and all of the number formats we might specify in the layout JSON.

There would be a straightforward rule for using it too:
1. If the axis's `format` isn't `"number"`, display the raw value and stop here. 
2. Start with `{notation: 'compact', compactDisplay: 'short'}`.
3. Merge the `numberFormat` object into that if it exists in the metric specification. 
4. Drop `minimumFractionDigits`, `minimumIntegerDigits`, and `minimumSignificantDigits`.
5. Pass those options into `intl.formatNumber`.

frontend/src/components/SectionLine/SectionLine.js
```diff
@@ -0,0 +1,22 @@
...
+
+const SectionLine = ({ layout, data }) => {
+  const xAxisKey = layout.x?.key
+  const yAxisKey = layout.y?.key
+  const chartData = data.map(item => ({
+    x: item[xAxisKey],
+    y: item[yAxisKey],
+  }))
+
+  return <LineChart data={chartData} title={layout.title} xAxisLabel="Population" yAxisLabel="Year" />
```
justinanderson, line 14: Could you alter detail_layout_example_small_business_support.json to include `title` values for the axis labels?
```json
{
  "title": "Projected Population",
  "key": "pop_proj",
  "type": "line",
  "x": {
    "key": "year",
    "title": "Year"
  },
  "y": {
    "key": "pop",
    "format": "number",
    "title": "Population"
  }
}
```

frontend/src/components/LineChart/LineChart.js
```diff
@@ -0,0 +1,53 @@
...
+  const theme = useTheme()
+  const domain = useMemo(() => getLineChartDomain(data, LINE_CHART_RATIO), [data])
+  const victoryTheme = useVictoryTheme(theme)
+
+  const ref = useRef()
+  const size = useClientSize(ref)
+
+  return (
+    <Box ref={ref} sx={{ columns: 1, marginTop: '-50px' }}>
+      <VictoryChart {...size} theme={victoryTheme}>
```
justinanderson, line 23: Could you try wrapping this in a [VictoryVoronoiContainer](https://formidable.com/open-source/victory/docs/victory-voronoi-container)? It would be nice if users didn't need to mouse precisely over a data point to see its tooltip.

frontend/src/components/LineChart/LineChart.js
```diff
@@ -0,0 +1,53 @@
...
+const LineChart = ({ data, title, xAxisLabel, yAxisLabel }) => {
+  const theme = useTheme()
+  const domain = useMemo(() => getLineChartDomain(data, LINE_CHART_RATIO), [data])
+  const victoryTheme = useVictoryTheme(theme)
+
+  const ref = useRef()
+  const size = useClientSize(ref)
+
+  return (
+    <Box ref={ref} sx={{ columns: 1, marginTop: '-50px' }}>
```
justinanderson, line 22: Why are `columns` and `marginTop` needed here?

frontend/src/components/LineChart/LineChart.js
```diff
@@ -0,0 +1,53 @@
...
+      <VictoryChart {...size} theme={victoryTheme}>
+        <VictoryAxis />
+        <VictoryAxis dependentAxis domain={domain} tickFormat={t => nFormatter(t)} />
+        <VictoryLine data={data} />
+        <VictoryScatter
+          data={data}
+          labels={() => ''}
+          labelComponent={
+            <VictoryTooltip
+              flyoutComponent={<ChartFlyOut title={title} xAxisLabel={xAxisLabel} yAxisLabel={yAxisLabel} />}
```
justinanderson, line 32: These tooltips are getting clipped, possibly by the column styling in SmallBusinessSupport. Screen Shot 2022-02-28 at 6 58 28 PMScreen Shot 2022-02-28 at 6 58 55 PM

justinanderson (changes requested):

frontend/src/components/LineChart/LineChart.js
leo-esaki: specifying `columns` was required because tooltip pane is displayed in the right column assuming it has 2 columns inheriting parent value.

frontend/src/components/LineChart/LineChart.js
```diff
@@ -1,38 +1,69 @@
...
+        domainPadding={{
+          x: [0, 15],
+          y: [0, 20],
+        }}
+        containerComponent={
+          <VictoryVoronoiContainer
+            labels={() => ' '}
+            labelComponent={
+              <VictoryTooltip
+                constrainToVisibleArea
```
justinanderson, line 57: It looks like the tooltip clipping issue I mentioned in my last review only happens in Safari and is related to the tooltip's DOM element sitting inside the CSS `columns` hierarchy. Safari appears to use an optimization for `columns` where it renders one long column texture and then wraps the result. If I switch to a 3 column layout, I can even make the [absolutely positioned tooltip wrap across columns](https://user-images.githubusercontent.com/244952/156217795-1e707c9f-7274-43d1-ad45-5865d2a46cb7.png).

`constrainToVisibleArea` seems to have no effect on the tooltip's positioning or clipping. Maybe it only works on tooltips without custom components?

The only solutions I can think of are:
1. Find another CSS method for wrapping content across 2 columns besides the `columns` property. That might be `grid`, but in my experience `grid` needs a predefined column height to know when to wrap.
2. Use a portal to opt the tooltip out of the `columns` hierarchy, making the tooltip's DOM element a child of the body element and calculate its position manually. There are a couple of examples of this in Victory's GitHub issues: [example one](https://github.com/FormidableLabs/victory/issues/576#issuecomment-342299341) and [example two](https://github.com/FormidableLabs/victory/issues/576#issuecomment-392499716).

frontend/src/components/LineChart/LineChart.js
```diff
@@ -1,38 +1,69 @@
...
+    },
+    [formatNumber, yAxis],
+  )
+
   return (
-    <Box ref={ref} sx={{ columns: 1, marginTop: '-50px' }}>
-      <VictoryChart {...size} theme={victoryTheme}>
-        <VictoryAxis />
-        <VictoryAxis dependentAxis domain={domain} tickFormat={t => nFormatter(t)} />
+    <Box ref={ref} sx={{ columns: 1, '& svg': { overflow: 'visible', position: 'relative', zIndex: 1 } }}>
```
justinanderson, line 44: Could you add a comment in code here for posterity, explaining why `columns: 1` was needed?

frontend/src/components/LineChart/LineChart.js
```diff
@@ -1,38 +1,69 @@
...
+    },
+    [formatNumber, xAxis],
+  )
+
+  const handleYTickFormat = useCallback(
+    t => {
+      return formatTickNumber(t, formatNumber, yAxis)
+    },
+    [formatNumber, yAxis],
+  )
```
justinanderson, line 41: No way to reduce these into a single `handleTickFormat(axis)`? Or is that not how `useCallback` can work?

frontend/src/utils/victory.js
```diff
@@ -3,15 +3,26 @@ import { useCallback, useEffect, useMemo, useState } from 'react'
...
   const max = Math.max(...data.map(item => item.y))
   const min = Math.min(0, Math.min(...data.map(item => item.y)))
-  const margin = (max - min) * 0.05
-  return [min, max + margin]
+  return [min, max]
+}
+
+export const formatTickNumber = (value, formatNumber, axis) => {
+  if (axis.format === 'number') {
+    const { maximumFractionDigits, minimumIntegerDigits, minimumSignificantDigits, ...numberFormat } =
```
justinanderson, line 11: Should be `minimumFractionDigits` instead of `maximumFractionDigits`. The goal is to remove any options which could make compactDisplay longer than expected.

justinanderson (changes requested):

frontend/src/components/LineChart/LineChart.js
leo-esaki: That will cause prop changes every time. Better to leave it as is now.

Issue #200, opened 2022-02-24
Add chloropleth map


justinanderson:
When a user browses a `map` type dataset, like when visiting /explore/small-business, they should see an edge-to-edge map. See [Figma mockup](https://www.figma.com/file/yaKUmuZM9xPQ1Qq5YVf5Yu/Design---Skills?node-id=2623%3A27784).

Like the detail page, the explore page's layout will be defined by a JSON response from the backend API. A example of that can be found at [explore_layout_small_business_support.json](https://github.com/CxSci/SA-dashboard/blob/master/frontend/src/mocked_api_responses/explore_layout_small_business_support.json).

See https://gist.github.com/justinanderson/d16aabc5551696997553c3bab66ef1f6 for an example of how the GDP Growth Potential chloropleth from explore_layout_small_business_support.json should end up rendering.

PR #204, opened 2022-03-03
Sa #200 chloropleth map

torosyaneric:
Issue #200


frontend/src/utils/helpers.js
```diff
@@ -0,0 +1,65 @@
...
+        width: layer?.paint?.default?.outline?.width,
+        opacity: layer?.paint?.default?.outline?.opacity ?? 1.0,
+      },
+      hover: {
+        width: layer?.paint?.hover?.outline?.width ?? 1.0,
+        opacity: layer?.paint?.hover?.outline?.opacity ?? 1.0,
+      },
+    },
+  }
+}
```
justinanderson, line 65: Please don't reformat the map layout object into the format I happened to use in my chloropleth example. My expectation was that you would write new code which generates Mapbox paint styles from the new format. This function shouldn't exist.

frontend/src/components/SimpleRange/SimpleRange.js
```diff
@@ -2,38 +2,46 @@ import { styled } from '@mui/material/styles'
...
+        content: '""',
+        display: 'block',
+        width: '100%',
+        height: '100%',
+        borderRadius: theme.shape.borderRadius,
+        transition: 'transform .4s linear',
+        transform: `translateX(${100 - value}%)`,
+        background: `linear-gradient(270deg, ${colorScheme?.[0] || theme.palette.canary.main} 0%, ${
+          colorScheme?.[1] || theme.palette.chestnutRose.main
+        } 50%, ${colorScheme?.[2] || theme.palette.ultramarine.main} 100%)`, // eslint-disable-line
```
justinanderson, line 22: 1. Instead of reversing the order of colorScheme in Map.js, why not apply them in reverse order here? Or rotate the gradient 180° and swap the order of the default colors?

2. There could be more or fewer than 3 colors, and their stops might not be evenly spaced. Like in `buildColorExpression`, the metric's `domain` and its `colorScheme` need to be interwoven to create the correct gradient. This domain's colors happen to be evenly distributed:
    ```json
    {
      "metric": {
        "domain": [0, 0.6, 1.2]
      },
      "paint": {
        "default": {
          "fill": {
            "colorScheme": ["#fff5eb", "#fd8d3c", "#7f2704"],
          },
        }
      }
    }
    ```
    And become:
    ```css
    background: linear-gradient(90deg, #fff5eb 0%, #fd8d3c 50%, #7f2704 100%);
    ```
    But this domain's colors are squeezed at the ends:
    ```json
    {
      "metric": {
        "domain": [0, 2, 8, 10]
      },
      "paint": {
        "default": {
          "fill": {
            "colorScheme": ["#00ffff", "#ffff00", "#ff0000", "#ff00ff"],
          },
        }
      }
    }
    ```
    And become:
    ```css
    background: linear-gradient(90deg, #00ffff 0%, #ffff00 20%, #ff0000 80%, #ff00ff 100%);
    ```

frontend/src/components/SimpleRange/SimpleRange.js
```diff
@@ -2,38 +2,46 @@ import { styled } from '@mui/material/styles'
...
-const SimpleRange = ({ min, max, value, style, color }) => {
+const SimpleRange = ({ min, max, value, style, color, colorScheme, height, sx = {}, ...otherProps }) => {
   const Range = style === 'gradient' ? GradientProgress : LinearProgress
   const percentage = Math.min(1, (value - min) / (max - min || 1))
   return (
     <Range
       value={percentage * 100}
       sx={{
-        height: 24,
+        height: height || 24,
```
justinanderson, line 35: Instead of supporting arbitrary heights, I'd prefer that SimpleRange take a `size` prop with `size='medium'` giving a height of 24 and `size='small'` giving a height of 14.

justinanderson (changes requested):

frontend/src/components/Map/Map.js
```diff
@@ -0,0 +1,397 @@
...
+
+  [`.${popupClassName}.immobile .mapboxgl-popup-content`]: {
+    background: 'rgba(255, 255, 255, 1.0)',
+  },
+})
+
+function Map({ config }) {
+  const mapContainer = useRef(null)
+  const map = useRef(null)
+  const clearMapFunctionsList = useRef([])
```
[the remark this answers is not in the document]
torosyaneric: Agree for clearMapFunctionsList: transformed it to use plain IDs
For map it's useful to keep it as ref so it won't rerun/update all effects and callbacks dependant on it

frontend/src/components/Map/Map.js
```diff
@@ -0,0 +1,397 @@
...
+
+    return () => {
+      map.current?.remove?.()
+      map.current = null
+    }
+  }, [initMap])
+
+  useEffect(() => {
+    getData()
+  }, [getData])
```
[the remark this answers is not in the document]
torosyaneric: This approach feels cleaner to me, no functional changes. Do you prefer it to be in a single effect?

frontend/src/components/Map/Map.js
```diff
@@ -0,0 +1,397 @@
...
+
+    // TODO: use client instance instead
+    try {
+      const res = await fetch(metricData?.data?.url)
+      const json = await res.json()
+      setData(json)
+    } catch (e) {
+      console.error(e)
+    }
+  }, [metricData?.data?.url])
```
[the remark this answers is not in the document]
torosyaneric: **getData** should be using useCallback as it's a dependency for that effect, and causes infinite loop when using arrow function as it's being recreated on every render.
**Memoizing** of vars is done for the same reason, just to prevent updating of effects and callbacks

frontend/src/components/Map/Map.js
```diff
@@ -0,0 +1,397 @@
...
+      const fillsId = `regions-${sourceLayer}-fills`
+      const linesId = `regions-${sourceLayer}-lines`
+      const hoverId = `regions-${sourceLayer}-hover-outline`
+
+      const mapConfig = createMapConfig({ metric: metricData, layer })
+
+      // TODO: temp - add other types
+      if (mapConfig.type !== 'chloropleth') return
+
+      map.current.addSource(sourceLayer, geometry)
```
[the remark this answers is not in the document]
torosyaneric: Sorry, didn't get this one: Which of these should be changed?
Isn't sourceLayer passed from the API response **layer.sourceLayer** ?
Where should I change those values?

frontend/src/components/Map/Map.js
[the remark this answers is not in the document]
justinanderson: It should still be multiple effects. I meant that effects whose entire bodies are lifted out into separate functions strikes me as unnecessary — especially for short functions like `initMap`. I could see an argument for lifting the chloropleth-specific part of `updateMap` into its own function, as `updateMap` will have multiple map types.

justinanderson (changes requested):

justinanderson (changes requested):
Looking better, thanks.

In addition to remaining unresolved comments:

- Please let me be the one to resolve PR conversations. They help me keep track of what to review, and it can be confusing if they're already marked as resolved.

- The merge commits in this branch could have been rebases. Per the [contribution guidelines](https://gist.github.com/justinanderson/0ff53d7eda9e4f4bf66c85683e28894f) I shared when you started, please rebase your branches rather than merging to keep your PRs up to date with master.

- When you add commits to address review comments, please break them into multiple commits when it makes sense to do so, and give them more descriptive commit messages. For example, the last commit could have been a few logically separate commits.

frontend/src/components/MapPopupContent/MapPopupContent.js
```diff
@@ -1,62 +1,69 @@
...
-      borderRadius={'10px'}
-      width={200}
-      boxShadow={'0px 2px 4px rgba(0, 0, 0, 0.25)'}>
-      <Typography fontSize={18} fontWeight={500} color={'#000'} mb={'8px'}>
-        {title}
-      </Typography>
-      <Typography fontSize={14} mb={0.5}>
-        {metricName}
-      </Typography>
+    <ThemeProvider theme={theme}>
```
justinanderson, line 14: Adding a new ThemeProvider for MapPopupContent is unnecessary. The underlying problem is the use of `ReactDOM.render` in Map.js, which has another undesirable side effect beyond disconnecting the popup from the main ThemeProvider. It leaks components. A new MapPopupContent is created every time `hoveredFeatureId` changes or a popup is expanded. Using a Portal would solve both of those problems and obviate the need for a second ThemeProvider here.

It's easier to implement than explain, so I'll push a commit which does it shortly.

justinanderson:
@torosyaneric Could you review my commits?

frontend/src/components/Map/Map.js
```diff
@@ -0,0 +1,467 @@
...
+  return (
+    <Box position={'absolute'} top={0} bottom={0} left={0} right={0}>
+      <GlobalStyles styles={mapGlobalStyles} />
+
+      <div ref={mapContainerRef} style={{ height: '100%', width: '100%' }} />
+      <div ref={popupContainerRef} />
+      <Portal container={popupContainerRef.current}>
+        <MapPopupContent {...popupData} />
+      </Portal>
+
```
torosyaneric, line 387: @justinanderson I think we don't need a portal as we can just put MapPopupContent inside popupContainerRef div:
```suggestion
      <div ref={popupContainerRef}>
        <MapPopupContent {...popupData} />
      </div>
```

torosyaneric:
@justinanderson reviewed your commits. Thank you, everything is cool, have only one suggestion for popups.

frontend/src/components/Map/Map.js
justinanderson: Thanks, you're right.

justinanderson (approved):

PR #205, opened 2022-03-04
SA-192 Implement show more/less feature with ExpandableContainer

leo-esaki:
Related to #192

### DIFF
--- a/frontend/src/components/ExpandableContainer/ExpandableContainer.js
+++ b/frontend/src/components/ExpandableContainer/ExpandableContainer.js
@@ -1,0 +1,42 @@
+import Box from '@mui/material/Box'
+import Link from '@mui/material/Link'
+import PropTypes from 'prop-types'
+import { useCallback, useMemo, useState } from 'react'
+
+import { EXPANDABLE_MINIMUM_ITEMS } from 'utils/constants'
+
+const ExpandableContainer = ({ data, children }) => {
+  const [expanded, setExpanded] = useState(false)
+  const expandable = data.length > EXPANDABLE_MINIMUM_ITEMS
+
+  const truncatedData = useMemo(
+    () => (expanded || !expandable ? data : data.slice(0, EXPANDABLE_MINIMUM_ITEMS)),
+    [data, expanded, expandable],
+  )
+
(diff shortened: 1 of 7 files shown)


justinanderson (changes requested):
In Safari, a metric's height doesn't shrink after clicking "Show Less". Screenshots [before](https://user-images.githubusercontent.com/244952/156708136-4c406f67-14dd-4684-910f-59848f0b89a6.png), [during](https://user-images.githubusercontent.com/244952/156708133-124d6db9-e98a-4100-84a0-2cdc5bf67944.png), after [after](https://user-images.githubusercontent.com/244952/156708130-4ea6712d-5898-4649-83a6-0f62c8a31f4f.png). This doesn't happen when the window is narrowed to a single column.

Also, you can see in that [during](https://user-images.githubusercontent.com/244952/156708133-124d6db9-e98a-4100-84a0-2cdc5bf67944.png) picture that Safari's `columns` implementation wraps the `margin-bottom` from the first column's section to the top of the second, pushing "Count" down slightly.

On Firefox, the metrics are wrapping to across columns. Note the weirdness in [Financial Transactions](https://user-images.githubusercontent.com/244952/156709106-008a2d71-dd2b-4553-85a1-9650c1f5fc78.png) and [Business Rental Costs](https://user-images.githubusercontent.com/244952/156709109-84fb9293-15f4-4d61-9445-9d0a41ec17d9.png). I'm guessing Firefox just needs more `break-inside: avoid`.

Everything seems fine in Chrome.

Safari's `columns` support has been troublesome enough, I'm wondering if we should try a different approach. What do you think about hardcoding columns in the layout JSON, and using a simple CSS `grid` layout instead of `columns`? It might be nice to be able to specify which column each metric appears in anyhow.

So instead of:
```js
"metrics": [
  "one",
  "two",
  "three",
  "four"
]
```

We would do something like:
```js
"metricColumns": [
  [
    "one",
    "two",
  ],
  [
    "three",
    "four"
  ]
]
```

justinanderson:
Merging this, as the CSS columns weirdness is a separate issue which shouldn't block this PR.

PR #206, opened 2022-03-04
SA-192 Re-implement columns layout with MetricsContainer

leo-esaki:
Related to #192


justinanderson (changes requested):
I think the `sx={{ columns: 1 }}` on LineChart and StackChart can be removed now. Other than that, this is great.

justinanderson (approved):

PR #208, opened 2022-03-07
SA-192 Implement select filter

leo-esaki:
Related to #192


justinanderson (approved):
Looks good to me. The styling will probably need to change at some point, but this is a good starting point.

Issue #209, opened 2022-03-09
Refine behavior of map popup


justinanderson:
This wasn't possible to show in the Figma mockup, but the map popup should have a specific behavior in response to various mouse events. The current behavior is close to what's desired.

When I say "feature" below, I mean in the Mapbox event sense where a feature is one of the polygonal map regions in `e.features`.

- [x] When the mouse is over a feature and the popup is not expanded, the popup should be visible, update it contents to match the hovered feature, and `follow` the mouse, staying just below the cursor.

- [x] After the mouse has been idle over a feature for 700ms, the popup should become expanded.

- [ ] The mouse should not be considered idle if the map is currently being panned, zoomed, rotated, or pitched, i.e. `map.isMoving()`.

- [x] When the popup is expanded, it should stop following the mouse, and its contents should become clickable and selectable.

- [ ] When the user clicks on a feature, the popup should immediately expand.

- [x] When the mouse is no longer over any feature and the popup is not expanded, the popup should not be visible.

- [ ] When the popup is expanded and the mouse moves more than 20px away from the popup's top or 10px away from its other sides, the popup should stop being expanded and go back to following the mouse and only being visible if the mouse is over a feature.

PR #211, opened 2022-03-10
Popup enhancements: #209

torosyaneric:
Not consider popup idle when map is moving,
Hide map popup when cursor is away,
Expand map popup on feature click

### DIFF
--- a/frontend/src/components/Map/Map.js
+++ b/frontend/src/components/Map/Map.js
@@ -5,7 +5,7 @@ import { InfoOutlined } from '@mui/icons-material'
 import SimpleRange from '../SimpleRange'
 import Select from '../Select'
 import MapPopupContent from '../MapPopupContent'
-import { MAPBOX_API_KEY } from '../../utils/constants'
+import { MAPBOX_API_KEY, MAP_POPUP_HORIZONTAL_DISTANCE, MAP_POPUP_VERTICAL_DISTANCE } from '../../utils/constants'
 
 mapboxgl.accessToken = MAPBOX_API_KEY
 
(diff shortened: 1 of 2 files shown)


justinanderson (changes requested):
Mostly right, but the map is still listening to mousemove events over regions near the popup. See in the video below how the popup resets when the cursor moves down even though it's expanded.

It might be easiest to set a transparent CSS padding buffer around the popup like `.mapboxgl-popup-content: {padding: '20 10 10 10'}` and adjust hoverPopup's offset to account for it. Then the padding works as a mouse event dead zone. Might be able to do without `onAllMapMouseMove` then.

Also, the popup's dead zone/buffer/padding should be 20 on top and 10 on the other sides, not 20 on top and bottom and 10 on the left and right.

https://user-images.githubusercontent.com/244952/157971796-1f09abd3-d41d-42c2-adac-d7a090c5cf81.mp4

torosyaneric:
@justinanderson fixed with the usage of onAllMapMouseMove. Padding approach didn't work that easy as we needed to keep popup following when moving mouse down, more changes were needed.

frontend/src/components/Map/Map.js
```diff
@@ -102,7 +107,7 @@ function Map({ config, hidePopup }) {
 
       function onMouseMove(e) {
         if (e?.features?.length > 0) {
-          if (hoveredFeatureId !== e.features[0].id) {
+          if (hoveredFeatureId !== e.features[0].id && !popupExpanded) {
```
justinanderson, line 110: I think you could move this `&& !popupExpanded` up one line to the `if (e?.features?.length > 0) {` and remove the `if (!popupExpanded) {` check below, since none of this block wants to run if the popup's already expanded.

justinanderson (commented):
One small code simplification suggested, but otherwise looks good.

justinanderson (approved):

Issue #207, opened 2022-03-04
Add scroll nav to detail page


justinanderson:
Detail pages should have a scroll nav — a list of links to each section which highlights the topmost visible section.

![image](https://user-images.githubusercontent.com/244952/156828220-6b8e6cf7-f924-41ab-b436-d7c78727d9b8.png)

![image](https://user-images.githubusercontent.com/244952/156828264-c497f488-7551-4fa9-a3fc-3892ccf8504f.png)

For a rough example of how this should behave, see the menu (Popular Items, Today's specials, etc) on [this food delivery site](https://www.doordash.com/store/don-julio-restaurant-somerville-522527/?pickup=true).

1. Scrolls with the page but sticks to the navbar.
2. Highlighted link changes as the page scrolls.
3. Scroll nav overflows and scrolls horizontally.
4. Shows left/right chevrons when overflowed in that direction. Clicking a chevron scrolls the list one page at a time (an amount equal to its own width).
5. Clicking a link jumps to the linked section. It's preferable that this is done with URLs hashes, so that people can share links to a particular section of a detail page like /explore/small-business/401011001#growth. This will probably require adding an `id` property to each section in the layout JSON response.

PR #212, opened 2022-03-11
SA-207 Implement smooth scroll navigation

leo-esaki:
Related to #207


frontend/src/components/SectionNavs/SectionNavs.js
```diff
@@ -0,0 +1,64 @@
...
+    const hash = event.target.href.split('#')[1]
+    const elem = document.getElementById(hash)
+    elem.scrollIntoView({ behavior: 'smooth' })
+    setLocationHash(`#${hash}`)
+  }, [])
+
+  useScrollSpy({ onScroll: handleScroll })
+
+  return (
+    <Box sx={{ position: 'fixed', width: '100%', zIndex: 'appBar' }}>
```
justinanderson, line 30: Should be something like `position: 'sticky', top: theme.scrolledHeaderHeight`.

frontend/src/routes/SmallBusinessSupport/SmallBusinessSupport.js
```diff
@@ -24,8 +26,9 @@ const SmallBusinessSupport = () => {
   const { sections } = sectionsLayout
   return (
     <Container sx={{ mt: 3 }}>
+      <SectionNavs sections={sections} />
```
justinanderson, line 29: To make the stickiness of the scrollspy easier to test, add in a placeholder for the content that will eventually be at the top of the page:

```suggestion
      <Typography variant="h1" gutterBottom>
        Adelaide
      </Typography>
      <Box sx={{height: 400, backgroundColor: 'lightGray'}} mb={4} />
      <SectionNavs sections={sections} />
```

justinanderson (changes requested):

frontend/src/components/SectionNavs/SectionNavs.js
```diff
@@ -0,0 +1,64 @@
+import Box from '@mui/material/Box'
+import Link from '@mui/material/Link'
+import { useCallback, useState } from 'react'
+import { useLocation } from 'react-router-dom'
+
+import { slugify } from 'utils/helpers'
+import useScrollSpy from 'utils/scrollSpy'
+
+export const SectionNavs = ({ sections }) => {
```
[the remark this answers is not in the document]
leo-esaki: Renamed to SectionNavbar

frontend/src/components/SectionNavs/SectionNavs.js
```diff
@@ -0,0 +1,64 @@
...
+  const location = useLocation()
+  const [locationHash, setLocationHash] = useState(location.hash)
+
+  const handleScroll = useCallback(element => {
+    const hash = `#${element.id}`
+    window.history.pushState(null, null, hash)
+    setLocationHash(hash)
+  }, [])
```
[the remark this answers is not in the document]
leo-esaki: Hash change only works for section wise. and I think we can use the current functionality.

frontend/src/components/SectionNavs/SectionNavs.js
leo-esaki: We can't just give the scrolledHeaderHeight, because this sticky navbar is going to be used in other pages with different header height. We should use a kind of React context to get the header height of the current page.

justinanderson: Works for me. 👍

frontend/src/components/SimpleCarousel/SimpleCarousel.js
```diff
@@ -0,0 +1,127 @@
...
+
+  return (
+    <Grid container alignItems="center">
+      {showArrows && (
+        <Grid item>
+          <IconButton onClick={handleLeft} disabled={scrollPos === 0}>
+            <ArrowLeft fontSize="small" />
+          </IconButton>
+        </Grid>
+      )}
```
justinanderson, line 86: Could you make the left and right <> buttons disappear when disabled?

justinanderson (changes requested):

frontend/src/components/SimpleCarousel/SimpleCarousel.js
```diff
@@ -0,0 +1,127 @@
...
+      <Grid
+        item
+        xs
+        sx={{ position: 'relative', overflow: 'hidden' }}
```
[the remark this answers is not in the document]
leo-esaki: There's a reason why I used left instead of scrolling
.
Browser doesn't support simultaneous smooth scroll in two elements.

frontend/src/components/SimpleCarousel/SimpleCarousel.js
leo-esaki: I also tried to do so, but if we make them disappear, the container width grows by the 2 arrow widths, and it affects the scroll effect. I tried to make them have absolute position and that requires to have white background for the arrows, but the problem is the header background has opacity value, and adding another half-transparent opacity will make it solid white background.

justinanderson: I'm ok with making the background opaque to make this easier. Absolute positioning over the ends of the scroll container would simplify things.

frontend/src/components/SimpleCarousel/SimpleCarousel.js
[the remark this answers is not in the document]
justinanderson: Being able to scroll the scrollnav is important for usability. I guess we'll have to animate the scrollNav manually.

justinanderson (changes requested):
- Change SimpleCarousel.js:96 to `{showArrows && scrollPos > 0 && (` so the left arrow doesn't reappear when Safari rubberbands/overscrolls on the left.

- Could you make it so the first navlink is active when the page first loads if there's no `location.hash`? Feels like we should make the first link active if we're scrolled above it and the last link active if we're scrolled below it.

- I found another edge case. Happens in Chrome, Safari, and Firefox.
  1. Start with a fresh, narrow page.
  2. Click on the "Demographics" navlink.
  3. Scroll the carousel all the way to the right.
  4. Click on the "Business Rental Costs" navlink.
  5. The carousel scrolls itself back to 0 unexpectedly.
  The same thing happens in reverse as well, starting from the bottom of the page and selecting "Demographics"

justinanderson (approved):

PR #217, opened 2022-03-16
Ignore mapbox-gl transpiling

### LINKED ISSUE #213: Address Mapbox transpilation errors in production builds
Production builds are running into a known transpilation issue with Mapbox. Reproduce with:
```sh
docker-compose -f docker-compose-prod.yml up
open http://localhost/explore/small-business/
```

See https://docs.mapbox.com/mapbox-gl-js/guides/install/#transpiling and https://github.com/mapbox/mapbox-gl-js/issues/10565 for details. I addressed this in Atlas v1 with https://github.com/CxSci/SA-dashboard/commit/3ec9ccd0d574d31a299cd73024e5484f4f991916, but there are better solutions than simply not transpiling Mapbox.

torosyaneric:
Resolves #213

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -65,7 +65,7 @@
   },
   "browserslist": {
     "production": [
-      ">0.2%",
+      ">0.2% and supports es6-class",
       "not dead",
       "not ie 11",
       "not chrome < 51",


torosyaneric:
@justinanderson I think not transpiling Mapbox is still better solution than transpiling Web Worker separately, as they say it significantly increases bundle size and negatively impacts rendering -  is only considered if targeting legacy browsers.

> If your application requires ES5 compatibility, then your module bundler needs to be configured to load and transpile Mapbox GL JS's Web Worker separately. This comes at the cost of significantly increasing the bundle size and negatively impacting rendering performance and you should only do this if you have a strong need for supporting legacy browsers. Mapbox GL JS can be configured with bundler specific worker-loader plugins.

justinanderson:
@torosyaneric But do we require ES5 compatibility? How does this compare with Mapbox's first suggestion to use browserlist?

torosyaneric:
@justinanderson it seems the browserslist approach wasn't working as our browserslist value is identical to their suggestion (+ "not op_mini all")

torosyaneric:
@justinanderson sorry, missed that one, fixed and pushed.

justinanderson:
@torosyaneric But, I mean, what's the difference between the two options? I don't have much experience with transpilation concerns and would like to understand the choice we're making.

torosyaneric:
@justinanderson I'm not 100% percent sure but I suppose using browserslist approach prevents the transpilation for whole codebase by rule "supports es6-class" and "import from !mapbox" was just ignoring the mapbox transpilation. I don't think these have advantages in terms of bundle size, maybe in build time. 
Do you want me to investigate this further?

justinanderson:
Force-pushed to remove noop commit and allow this branch to rebase cleanly.

PR #240, opened 2022-03-31
Make MUI links use react-router Link component

### LINKED ISSUE #234: Use react-router with MUI Link for all links
- `<Link>`s in AtlasBreadcrumbs aren't react-router links, so using them causes a full site load.
- `<Link>`s in MapPopupContent aren't MUI links, so they have their own custom styling method which is disconnected from the overall site theme.

We should do something like https://mui.com/guides/routing/#global-theme-link, so we can use MUI themed links everywhere but with react-router behavior by default.

torosyaneric:
Resolves #234

### DIFF
--- a/frontend/src/components/MapPopupContent/MapPopupContent.js
+++ b/frontend/src/components/MapPopupContent/MapPopupContent.js
@@ -1,7 +1,6 @@
 import React from 'react'
-import { Box, Button, styled, Typography } from '@mui/material'
+import { Box, Button, styled, Typography, Link } from '@mui/material'
 import { ChevronRight } from '@mui/icons-material'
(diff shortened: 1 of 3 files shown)


justinanderson:
Shouldn't this also replace uses of raw react-router `<Link>`s with MUI `<Link>`s, like in [Home](https://github.com/CxSci/SA-dashboard/blob/master/frontend/src/routes/Home/Home.js#L2) and [MapPopupContent](https://github.com/CxSci/SA-dashboard/blob/master/frontend/src/components/MapPopupContent/MapPopupContent.js#L4)?

justinanderson (approved):

justinanderson:
Rebased onto master to resolve conflicts

PR #244, opened 2022-04-01
Use redux state instead of mock data of dataset map

leo-esaki:


justinanderson (changes requested):
The "Adelaide" in SmallBusinessSupport.js should be coming from `_atlas_title`.

justinanderson (approved):

PR #247, opened 2022-04-02
Sa 232 search functionality

### LINKED ISSUE #232: Hook up search field to search endpoint
The search field on pages like http://localhost:3000/explore/small-business-support should query the backend's search endpoint, `/datasets/{dataset}/search/?q={query}`, and work similarly to Atlas v1.

Example searches:
"adelaide" -> http://localhost:8000/datasets/small-business-support/search/?q=adelaide
"kangaroo island" -> http://localhost:8000/datasets/small-business-support/search/?q=kangaroo%20island

The appearance of the search result list can be left as is for now. This task is about interacting with the API and getting the right interaction between the search field and the map.

Atlas v1 behavior to be copied:

- Search should update live as a user types, with a 500ms hysteresis/debounce to wait for a user to stop typing before actually searching.
- Search results should have the concept of a highlighted row, displayed visually by shading the background of the row slightly. The first row should be highlighted whenever search results change, and the highlight should change when a user mouses over a different row or presses up/down arrow keys while the search field is focused.
- While a row is highlighted, a non-expanded popup should be shown over the related map feature. Popups shown this way should not expand after a timeout. The map shouldn't move to show map features which aren't currently visible.
- Search results should have the concept of a selected row, which happens when a user presses enter when the search field is focused and when a user clicks on a row.
- When a row is selected
    - The search field should lose focus
    - The search results should close
    - An expanded popup should appear on the related map feature
    - The map should pan and zoom to focus on the related map feature, probably using [map.fitBounds](https://github.com/CxSci/SA-dashboard/blob/master/frontend-old/src/components/map.js#L610-L622)
- The usual popup behavior from hovering the map should still occur while search results are visible.

torosyaneric:
Resolves #232


frontend/src/routes/Dataset/Dataset.js
```diff
@@ -23,6 +25,14 @@ function Dataset() {
   const datasetName = dataset?.title || ''
   const searchPlaceholder = datasetConfig?.searchPlaceholder || ''
 
+  const [selectedSearchResult, setSelectedSearchResult] = useState(null)
+  const [highlightedSearchResult, setHighlightedSearchResult] = useState(null)
+
+  const handleSearchChange = useCallback(e => {
+    // TODO: 'small-business' -> dataset?.id
```
torosyaneric: @justinanderson shall this endpoint be changed to "small-business-support"? **datasetId** has that value in this case

torosyaneric:
Deleting PR + branch as diffs are merged

PR #248, opened 2022-04-02
Sa 232 search functionality

### LINKED ISSUE #232: Hook up search field to search endpoint

torosyaneric:
Resolves #232

### DIFF
--- a/frontend/src/components/Header/SearchInput.js
+++ b/frontend/src/components/Header/SearchInput.js
@@ -1,12 +1,23 @@
-import React from 'react'
+import React, { useRef } from 'react'
+import { useSelector } from 'react-redux'
 import { Autocomplete, InputAdornment, TextField } from '@mui/material'
 import { Close as CloseIcon, Search as SearchIcon } from '@mui/icons-material'
+import { searchListSelector } from '../../store/modules/search'
 
-function SearchInput({ placeholder }) {
-  const options = [
-    { label: 'The Godfather', id: 1 },
-    { label: 'Pulp Fiction', id: 2 },
-  ]
+function SearchInput({ placeholder, onChange = () => null }) {
+  const changeTimeoutRef = useRef(0)
+
+  const options = useSelector(searchListSelector)
+  console.log(searchListSelector)
+  console.log(options)
+
+  const handleChange = e => {
+    clearTimeout(changeTimeoutRef.current)
+
+    changeTimeoutRef.current = setTimeout(() => {
+      onChange(e)
+    }, 500)
+  }
 
   return (
     <Autocomplete
@@ -20,6 +31,7 @@ function SearchInput({ placeholder }) {
           {...params}
           sx={{ width: theme => theme.components.searchInput.width }}
           variant={'filled'}
+          onChange={handleChange}
           InputProps={{
             ...params.InputProps,
             placeholder,
--- a/frontend/src/components/Header/SearchInput.js
+++ b/frontend/src/components/Header/SearchInput.js
@@ -1,7 +1,7 @@
 import React, { useCallback, useEffect, useState } from 'react'
 import { useSelector } from 'react-redux'
 import debounce from 'lodash/debounce'
-import { Autocomplete, CircularProgress, InputAdornment, Stack, TextField, Typography } from '@mui/material'
+import { Autocomplete, CircularProgress, IconButton, InputAdornment, Stack, TextField, Typography } from '@mui/material'
 import { Close as CloseIcon, Search as SearchIcon } from '@mui/icons-material'
 import { searchListSelector } from '../../store/modules/search'
 import { isRequestPending } from '../../store/modules/api'
@@ -88,8 +88,23 @@ function SearchInput({ placeholder, onChange = () => null, onSelect = () => null
                 <SearchIcon />
               </InputAdornment>
             ),
-            endAdornment: isLoading ? <CircularProgress color="inherit" size={20} /> : null,
-            sx: { paddingRight: '14px !important' },
+            endAdornment: (
+              <React.Fragment>
+                {isLoading ? <CircularProgress color="inherit" size={20} /> : null}
+
+                {inputValue && (
+                  <IconButton
+                    onClick={() => setInputValue('')}
+                    sx={{
+                      color: theme => theme.components.autocompleteClearIcon.color,
+                      position: 'absolute',
+                      right: '2px',
+                    }}>
+                    <CloseIcon />
+                  </IconButton>
+                )}
+              </React.Fragment>
+            ),
           }}
         />
       )}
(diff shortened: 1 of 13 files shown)


frontend/src/components/Map/Map.js
```diff
@@ -413,6 +416,49 @@ function Map({ config, hidePopup, datasetId }) {
...
+      })
+
+      hoverPopupRef.current.setLngLat({ lng: popupLng, lat: popupLat })
+      hoverPopupRef.current.setDOMContent(popupContainerRef.current).addTo(map.current)
+
+      if (expandPopup) {
+        hoverPopupRef.current.addClassName('immobile')
+      }
+    },
+    [data, metricConfig],
```
justinanderson, line 459: Missing hook dependencies `colorScheme` and `domain`.

frontend/src/routes/Dataset/Dataset.js
```diff
@@ -9,19 +9,31 @@ import Dashboard from 'components/Dashboard'
...
   const DataSetComponent = getDatasetComponent(data?.type)
   const datasetName = dataset?.title || ''
   const searchPlaceholder = data?.searchPlaceholder || ''
   const theme = useMemo(() => initTheme(data?.theme), [data?.theme])
 
+  const handleSearchChange = useCallback(
+    e => {
+      dispatch(getSearchList({ datasetId, params: { q: e?.target?.value } }))
+    },
+    [datasetId],
```
justinanderson, line 34: Missing hook dependency `dispatch`.

justinanderson (changes requested):

frontend/src/components/Header/SearchInput.js
```diff
@@ -1,25 +1,97 @@
...
+  )
+
+  const handleOptionMouseEnter = useCallback(
+    e => {
+      const featureId = e?.target?.dataset?.featureId
+      const option = options?.find(item => item?.id === featureId)
+      onHighlightChange(option)
+    },
+    [options, onHighlightChange],
+  )
```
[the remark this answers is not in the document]
torosyaneric: Weird, have already tried to use that prop but wasn't firing the event. But now it works

frontend/src/components/Header/SearchInput.js
```diff
@@ -1,96 +1,89 @@
...
+        if (debouncedFuncRef.current?.cancel) {
+          debouncedFuncRef.current.cancel()
+        }
+
+        onChange(e?.target?.value)
+      }
     },
-    [options, onHighlightChange],
+    [onChange],
   )
```
justinanderson, line 40: Now that this is further along, I'm getting a clearer idea of what behavior I'd like to see from the search field. I think it was the wrong idea to overload "Enter" to mean both "search now" and "select the highlighted option".

How does this sound for behavior?

- Don't do any special handling of key presses. Rely on `<Autocomplete>`'s default "Enter" behavior.

- When the input is focused
    - If there are options, show them.
    - If the input is blank, keep the options popup closed and clear the options list.
    - If the input is not blank and there are no results after its search query returns, show "No results".
    - I think these should be doable with a combination of `noOptionsText`, `onOpen`, `open`, and `openOnFocus`.

- When the input text changes
    - The autocomplete should show an indeterminate progress indicator on its right like https://mui.com/components/autocomplete/#load-on-open while it waits for results.
    - Unlike that example, it shouldn't replace the options with "Loading…" text. Previous options should remain while waiting for new options to load.

- When a user selects an option
    - The input text should remain as it is. It should not be cleared or set to the name of the selected option. This should make it easier for a user to revisit their last search after making a selection.

frontend/src/components/Header/SearchInput.js
```diff
@@ -1,96 +1,89 @@
...
+
+    debouncedFunc()
+    return () => {
+      debouncedFunc.cancel()
+    }
+  }, [inputValue, onChange])
 
   return (
     <Autocomplete
-      freeSolo
```
justinanderson, line 13: I think we should keep `freeSolo` and remove `value` from from the `<Autocomplete>`. We're using it as a search field rather than a select element, so we don't need it to remember which option was selected. I would like the field to retain any text a user put in there, however.

justinanderson (changes requested):

frontend/src/components/Header/SearchInput.js
torosyaneric: @justinanderson if we use freeSolo Mui doesn't show "no results" block. With current setup user can type anything and it's not being cleared nor replaced with selected value.

frontend/src/components/Header/SearchInput.js
justinanderson: I've added checkboxes to my previous comment to mark what parts aren't working right yet.

- The options popup shouldn't open or show "No results" if the input is empty.

- After applying the previous rule, the options popup should open whenever it's focused. It does if you click on the input but not if you tab to it.

- "No results" should only appear after a search has completed and returned no results.

justinanderson (changes requested):

frontend/src/components/Header/SearchInput.js
```diff
@@ -78,8 +88,23 @@ function SearchInput({ placeholder, onChange = () => null, onSelect = () => null
...
+                  <IconButton
+                    onClick={() => setInputValue('')}
+                    sx={{
+                      color: theme => theme.components.autocompleteClearIcon.color,
+                      position: 'absolute',
+                      right: '2px',
+                    }}>
+                    <CloseIcon />
+                  </IconButton>
+                )}
```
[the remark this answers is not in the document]
torosyaneric: Yes, I think it's better to override the `InputProps.endAdornment`

justinanderson (approved):

PR #253, opened 2022-04-06
Create a component to render a SVG map with GeoJSON data

leo-esaki:


frontend/src/components/StaticMap/StaticMap.js
```diff
@@ -0,0 +1,81 @@
+import React, { useEffect, useMemo, useRef, useState } from 'react'
+import { generateSize, getScaleTranslation } from './StaticMap.utils'
+import { geoMercator, geoPath } from 'd3-geo'
+import { useTheme } from '@mui/system'
+import Box from '@mui/material/Box'
+import PropTypes from 'prop-types'
+
+const StaticMap = ({ geoJSON, square, height: mapHeight, areaId }) => {
```
justinanderson, line 8: There's a way to use an SVG's `viewBox` and `preserveAspectRatio` to skip the need to define its size or worry about resize events, but this is performing well enough that I'd say let's not bother trying that unless we run into performance problems.

frontend/src/components/StaticMap/StaticMap.utils.js
```diff
@@ -0,0 +1,18 @@
...
+  }
+}
+
+export const scaleRatio = 0.55
+
+export const getScaleTranslation = (pathGenerator, location, width, height) => {
+  const b = pathGenerator.bounds(location)
+  const s = scaleRatio / Math.max((b[1][0] - b[0][0]) / width, (b[1][1] - b[0][1]) / height)
+  const t = [(width - s * (b[1][0] + b[0][0])) / 2, (height - s * (b[1][1] + b[0][1])) / 2]
+  return [s, t]
```
justinanderson, line 17: This could use a line comment explaining what it does and why 0.55 was chosen. It took me a little bit to realize `getScaleTranslation ` is what zooms to fit around the selected feature.

justinanderson (changes requested):
After a page load, the static maps stay blank until I resize the page. I see this in Safari, Chrome, and Firefox.

frontend/src/components/StaticMap/StaticMap.js
```diff
@@ -29,7 +29,7 @@ const StaticMap = ({ geoJSON, square, height: mapHeight, areaId }) => {
 
   useEffect(() => {
     const handleResize = () => setSize(generateSize(ref, square, mapHeight))
-    handleResize()
+    setTimeout(() => handleResize(), 0)
```
justinanderson, line 32: This fixes the "blank on first visit" issue on Firefox and Chrome, but it's still black most of the time on Safari. On Safari, `ref` is still null when the timeout fire. Raising the timeout delay to 1000ms makes it render correctly on Safari, but that's a fragile solution. I think the right solution is either to [use a callback ref](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node) in combination with the current event listener or to copy one of the many useSize/useDimension/useResize hooks out there.

justinanderson (changes requested):

justinanderson (approved):

PR #259, opened 2022-04-12
Show loading indication on detail page

leo-esaki:


justinanderson:
I never see any skeletons running this, even after adding `sleep(10)` to the [backend's detail endpoint](https://github.com/CxSci/SA-dashboard/blob/master/backend/api/views.py#L415).

I think [this check in SmallBusinessSupport.js](https://github.com/CxSci/SA-dashboard/blob/SA-249-show-loading-indication-on-detail-page/frontend/src/routes/SmallBusinessSupport/SmallBusinessSupport.js#L29-L31) should only short-circuit to a spinner when `sectionsLayout` is missing. If we know `sectionsLayout` but don't have `sectionsData`, then we know the layout and can show skeletons while we wait for data to show up. However, when I make that change locally I see downstream errors from components that weren't expecting null.

frontend/src/components/SectionSimpleBar/SectionSimpleBar.js
```diff
@@ -1,26 +1,43 @@
...
   const xAxisKey = layout.x?.key
   const yAxisKey = layout.y?.key
+  const ref = useRef()
+  const defaultContainerWidth = 500
+  const defaultContainerHeight = 250
+  const size = useClientSize(ref, defaultContainerWidth, defaultContainerHeight, 0.5)
+
+  if (!layout) {
+    return <Skeleton variant="rectangular" sx={{ borderRadius: 1 }} />
+  }
```
justinanderson, line 22: Individual metrics shouldn't have to worry about showing skeletons for `!layout`. If the layout is unknown, there's a full page spinner blocking any metrics from rendering.

justinanderson (changes requested):
- [ ] It would be good to have a skeleton for StaticMap as well.

- [ ] The square StaticMap and title which show up in the header when the page is scrolled should have skeletons too. Currently they're missing while waiting for data.

backend/api/views.py
```diff
@@ -416,6 +417,7 @@ def get(self, request, dataset=None, pk=None, format=None):
         dsn = "postgresql://{USER}:{PASSWORD}@{HOST}:{PORT}/{NAME}".format(
             **settings.DASHBOARD_DATABASE
         )
+        time.sleep(20)
```
justinanderson, line 420: This shouldn't be committed.

frontend/src/routes/SmallBusinessSupport/SmallBusinessSupport.js
```diff
@@ -36,7 +36,11 @@ const SmallBusinessSupport = ({ sectionsData, sectionsLayout, entryId, geoJSON }
         <Typography variant="h1" gutterBottom>
           {sectionsData ? sectionsData._atlas_title : <Skeleton />}
         </Typography>
-        <StaticMap height={400} square={false} geoJSON={geoJSON} areaId={entryId} />
+        {geoJSON ? (
+          <StaticMap height={400} square={false} geoJSON={geoJSON} areaId={entryId} />
+        ) : (
+          <Skeleton height={400} />
```
justinanderson, line 42: The default is `variant="text"`, which scales the height to 60%.
```suggestion
          <Skeleton height={400} variant="rectangular" />
```

frontend/src/routes/DatasetEntry/DatasetEntry.js
```diff
@@ -92,12 +93,14 @@ const DatasetEntry = () => {
           contentScrolled: {
             left: (
               <>
-                {sectionsData?._atlas_header_image && (
+                {sectionsData?._atlas_header_image ? (
```
justinanderson, line 96:
```suggestion
                {datasetGeoJSON ? (
```

PR #261, opened 2022-04-13
Use redux instead of fetch in Map

leo-esaki:


frontend/src/store/modules/dataset/sagas.js
```diff
@@ -37,9 +37,19 @@ const getDatasetGeoJSON = createApiCallSaga({
   allowedParamKeys: ['ids', 'include_neighbors', 'format'],
 })
 
+const getDatasetMapData = createApiCallSaga({
```
justinanderson, line 40: Better to name this `getDatasetMetricData`.

justinanderson (changes requested):

justinanderson (approved):

PR #265, opened 2022-04-18
Handle missing data in detail page gracefully

leo-esaki:


justinanderson (changes requested):
Looks good for http://localhost:3000/explore/small-business-support/497979799, but there are some issues with http://localhost:3000/explore/small-business-support/406011135.

- Runtime warning: ``Warning: Failed prop type: Invalid prop `domain` supplied to `VictoryAxis`.``

- `null` shouldn't be treated as `0`. 1BR houses under Residential Housing should say "No data" instead of "A$0.00".

- I also noticed that the `filters` on "Apartments: Median Weekly Rent" and "Houses: Median Weekly Rent" aren't being applied — they're both showing all `residential_housing_median` data rather than just their filtered `type`. I'll create a separate issue to track that. Don't feel obligated to fix it in this PR.

justinanderson (changes requested):
Pretty close, just a few styling nits:

- "No Data" should be "No data"

- "No data" should be styled the same as the metric would have been. I think this amounts to changing every instance of `<Typography>No Data</Typography>` to `<Typography variant="fieldValue">No data</Typography>`.

- Should have consistent spacing in `<SectionText>` when some items have no data. I think that's just a matter of adding `gutterBottom` to the `<Typography>` in the `<ExpandableContainer>` in `<SectionText>`.

justinanderson (approved):

PR #268, opened 2022-04-21
Add support multiple lines in LineChart component

leo-esaki:


frontend/src/components/LineChart/LineChart.js
```diff
@@ -61,9 +60,13 @@ const LineChart = ({ data, title, xAxis, yAxis }) => {
             }
           />
         }>
-        <VictoryAxis tickFormat={handleXTickFormat} />
-        <VictoryAxis dependentAxis domain={domain} tickFormat={handleYTickFormat} />
-        <VictoryLine data={data} />
+        <VictoryAxis fixLabelOverlap tickFormat={handleXTickFormat} />
+        <VictoryAxis dependentAxis tickFormat={handleYTickFormat} />
+        <VictoryStack colorScale={STACK_COLORS}>
```
justinanderson, line 65: `<LineChart>` isn't a stack. Each line should be drawn independently from the other. For instance, if I alter the layout for "Business Counts" to be a line chart, this point with a value of 0 should be down at the bottom and not up around 800.

image

justinanderson (changes requested):
In addition to the one line comment, let's remove the `<VictoryScatter>` from `<LineChart>`.  It's looking like most of the data we'll be adding to the project will make line charts too crowded to show both lines and circles.

justinanderson (approved):

PR #275, opened 2022-04-28
Set static height for LineChart and StackChart

leo-esaki:


justinanderson (changes requested):
Is there a way to do this so that charts still stretch horizontally as the window size changes?

There's also some odd scaling of the chart and its labels at some widths, making its features extra small. Looks like maybe the chart is scaling according to available width and then scaling itself down to fit the fixed height.

So this looks correct when the window is as narrow as I can make it:
Screen Shot 2022-04-28 at 2 36 21 PM

But the widest window with a single column scales the chart oddly.
Screen Shot 2022-04-28 at 2 35 56 PM

justinanderson (approved):

PR #278, opened 2022-05-02
Make map metric select work,

### LINKED ISSUE #272: Make map layer switcher work
Steps to reproduce:
1. Visit http://localhost:3000/explore/small-business-support
2. Use the `<Select>` at bottom left to switch from "GDP Growth Potential" to another metric like "Mean Income".

Instead of showing the new metric, the page crashes because the source id and layer ids of the new metric are the same as the old one. Old sources and layers should be removed when swapping metrics.

If the source is actually the same, it would be great to reuse it so we don't end up reloading the same data. Reusing similar layers is less important, as that'll be difficult to decide and not as big of a performance gain.

torosyaneric:
Update map cleanup logic,
Update SimpleRange component gradient generation function,

Resolves #272

### DIFF
--- a/frontend/src/components/Map/Map.js
+++ b/frontend/src/components/Map/Map.js
@@ -60,7 +60,6 @@ function Map({
   const [metricConfig, setMetricConfig] = useState(null)
   const [colorScheme, setColorScheme] = useState([])
   const [popupData, setPopupData] = useState(null)
-  const sourceIds = useRef([])
 
   const domain = useMemo(() => metricConfig?.layers?.[0]?.metric?.domain || [], [metricConfig?.layers])
 
@@ -264,6 +263,7 @@ function Map({
     }
 
     const layerIds = []
+    const sourceIds = []
 
     const { geometry, layers } = metricConfig || {}
     const mapType = metricConfig?.type
@@ -272,8 +272,9 @@ function Map({
     const metricKey = 'data'
     const sourceName = createSourceName(geometry)
 
-    if (!sourceIds.current.includes(sourceName)) {
-      sourceIds.current.push(sourceName)
+    sourceIds.push(sourceName)
+    const source = map.current.getSource(sourceName)
+    if (!source) {
       map.current.addSource(sourceName, geometry)
     }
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/Map/Map.js
```diff
@@ -57,15 +59,14 @@ function Map({
...
-  const [, setSourceIds] = useState([])
   const [popupData, setPopupData] = useState(null)
+  const sourceIds = useRef([])
 
   const domain = useMemo(() => metricConfig?.layers?.[0]?.metric?.domain || [], [metricConfig?.layers])
 
   const initPopup = useCallback(
     ({ foreignKey, metricKey, titleKey, layerId, sourceLayer }) => {
-      const sourceName = `source_${metricConfig?.id || ''}`
+      const sourceName = `source_${metricConfig?.geometry?.promoteId || ''}`
```
justinanderson, line 69: We might end up with two sources in the future with the same `promoteId` but a different `url`. Maybe make this a hash of `url` and `promoteId`? Or something else that varies with both of those variables.

Same goes for other lines which generate the source name in this file.

justinanderson (changes requested):
@torosyaneric Pretty cool to see layer switching so smooth and quick.

Just one small requested change.

frontend/src/components/Map/Map.js
torosyaneric: @justinanderson yes, I was wondering between url and promoteId but thought both will be unique. Will change to url hash

justinanderson (changes requested):
@torosyaneric I just realized there's a console error after switching metrics:
```
TypeError: this.expression.evaluate is not a function. (In 'this.expression.evaluate({
          zoom: 0
        }, r, n, void 0, i)', 'this.expression.evaluate' is undefined)
```

That it happens 3 times each time I change map metrics hints at a relation to the way we're adding and/or removing layers.

torosyaneric:
@justinanderson the only way I found to get rid of those errors is clearing the sources too

justinanderson:
@torosyaneric It looks like making the layer ids unique fixes it too, but only if the layer ids are different every time, e.g. including a timestamp in the layer id. Seems likely that this is related to https://github.com/mapbox/mapbox-gl-js/issues/9731, though this is about reusing layer ids when the layer type varies.

That points to a memory leak in Mapbox which we can only avoid by replacing the source every time we change layers. Annoying that we need to add that slowdown to the user experience, but it seems like the best choice while we wait for Mapbox to fix that issue.

justinanderson (approved):

PR #283, opened 2022-05-04
Truncate header breadcrumbs on smaller screens: rightmost items should be truncated first

### LINKED ISSUE #279: Truncate breadcrumbs when room is limited
When components in the header are fighting for space, breadcrumbs should be the first to give in.

- They shouldn't wrap.
- They should be replaced with ellipses "…" as needed.
- The rightmost breadcrumbs should be replaced first to make room.

Screen Shot 2022-05-03 at 7 20 28 PM
Screen Shot 2022-05-03 at 7 21 37 PM
Screen Shot 2022-05-03 at 7 24 21 PM

torosyaneric:
Resolves #279


frontend/src/components/AtlasBreadcrumbs/AtlasBreadcrumbs.js
```diff
@@ -1,20 +1,90 @@
...
+  useEffect(() => {
+    function handleResize() {
+      let breadcrumbsWidth = containerWidth
+      const windowWidth = window.innerWidth
+      const headerRightContentNode = document.querySelector('.header__right-content')
+      const headerActionBtnNode = document.querySelector('.header__action-btn')
+      const rightContentWidth = headerRightContentNode?.offsetWidth
+      const actionBtnWidth = headerActionBtnNode?.offsetWidth + 16
+
+      const allocatedBreadcrumbsWidth = windowWidth - rightContentWidth - actionBtnWidth - 40
```
justinanderson, line 55: It would be good to include comments explaining where `16` and `40` come from, so we know if and how to change them in the future.

justinanderson (changes requested):

frontend/src/utils/theme.js
```diff
@@ -135,8 +135,27 @@ const initTheme = mode => {
...
           li: {
+            whiteSpace: 'nowrap',
             fontWeight: 'bold',
+
+            '.header__link': {
+              '.header__link__ellipsis': {
+                display: 'none',
+              },
+
+              '&.header__link--truncated': {
```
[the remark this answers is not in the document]
torosyaneric: @justinanderson I've used [BEM](http://getbem.com/introduction/) naming. Do you want me to change to hyphens?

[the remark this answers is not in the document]
justinanderson: @torosyaneric No, it's good as is. I just hadn't heard of BEM before. I didn't know `__` and `--` had specific meanings.

frontend/src/components/AtlasBreadcrumbs/AtlasBreadcrumbs.js
```diff
@@ -1,20 +1,92 @@
...
+          truncateCount++
+        }
+      })
+
+      setTruncateCount(truncateCount)
+    }
+
+    handleResize()
+    window.addEventListener('resize', handleResize)
+
```
leo-esaki, line 74: Here's my suggestion:
We render 2 breadcrumbs one is for calculating the width of each item and its visibility will be hidden, along with aria-role set as none. that will remove the logic of manually cloning and adding DOMs to documents and setting css styles.
Also, using flexbox, the width of breadcrumb links area can be obtained just specifying a ref to the links container element. No need to calculate the width from left and right sections of the breadcrumbs.

frontend/src/components/AtlasBreadcrumbs/AtlasBreadcrumbs.js
```diff
@@ -1,20 +1,101 @@
...
+    function handleResize() {
+      let breadcrumbsWidthTemp = breadcrumbsWidth
+      const allocatedBreadcrumbsWidth = containerRef.current?.offsetWidth
+
+      let truncateCount = 0
+      links.forEach((item, index) => {
+        if (breadcrumbsWidthTemp > allocatedBreadcrumbsWidth) {
+          const itemWidth = widthsList[links.length - 1 - index]
+          const ellipsisWidth = 22
```
leo-esaki, line 76: Can you define a constant at the top of the component or in the .utils.js file?

leo-esaki (changes requested):

justinanderson (approved):

PR #286, opened 2022-05-05
Make header search input width depend on placeholder

### LINKED ISSUE #280: The search field shouldn't truncate its placeholder string
The Atlas' various datasets can have different search placeholder strings with varying lengths, but the search field is currently set to always be 250px wide which truncates longer placeholders:
Screen Shot 2022-05-03 at 7 47 59 PM

It'd be good to leave the field's minimum width at 250px but still allow it to grow to fit its placeholder without truncation:
Screen Shot 2022-05-03 at 7 45 50 PM
Screen Shot 2022-05-03 at 7 52 33 PM
Screen Shot 2022-05-03 at 7 58 15 PM

torosyaneric:
Resolves #280

### DIFF
--- a/frontend/src/components/Header/SearchInput.js
+++ b/frontend/src/components/Header/SearchInput.js
@@ -79,11 +79,14 @@ function SearchInput({ placeholder, onChange = () => null, onSelect = () => null
       renderInput={params => (
         <TextField
           {...params}
-          sx={{ width: theme => theme.components.searchInput.width }}
           variant={'filled'}
           onChange={event => setInputValue(event?.target?.value)}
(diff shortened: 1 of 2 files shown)


justinanderson (changes requested):
This doesn't seem to do the right thing.

Safari:
Screen Shot 2022-05-05 at 12 24 50 PM
Firefox:
Screen Shot 2022-05-05 at 12 24 59 PM
Chrome:
Screen Shot 2022-05-05 at 12 25 10 PM
Desired result:

justinanderson (approved):

PR #290, opened 2022-05-10
Fix map zoom performance on multiple metric changes: fix onMapMove function event listener removing

### LINKED ISSUE #282: Map becomes less responsive after switching metrics
The map becomes progressively slower to zoom as you switch metrics. It becomes obvious after switching 10 or so times.

torosyaneric:
Resolves #282

### DIFF
--- a/frontend/src/components/Map/Map.js
+++ b/frontend/src/components/Map/Map.js
@@ -45,6 +45,7 @@ function Map({
 }) {
   const mapContainerRef = useRef(null)
   const popupContainerRef = useRef(null)
+  const popupLayerIdRef = useRef(null)
   const hoverPopupRef = useRef(null)
   const hoveredFeatureIdRef = useRef(null)
   const map = useRef(null)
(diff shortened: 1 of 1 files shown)


frontend/src/components/Map/Map.js
```diff
@@ -239,7 +244,7 @@ function Map({
 
       // map click event
       if (onMapClickRef.current) {
-        map.current.off('mouseleave', layerId, onMapClickRef.current)
+        map.current.off('mouseleave', prevLayerId, onMapClickRef.current)
```
justinanderson, line 247: Shouldn't this be `'click'`, not `'mouseleave'`?

frontend/src/components/Map/Map.js
```diff
@@ -395,16 +402,23 @@ function Map({
...
+      map.current.off('mousemove', onAllMapMouseMoveRef.current)
+
       if (layerIds.length) {
         layerIds.forEach(id => {
           const layer = map.current.getLayer(id)
           if (layer) {
+            map.current.off('mousemove', id, onMouseMoveRef.current)
+            map.current.off('mouseleave', id, onMouseLeaveRef.current)
+            map.current.off('mouseleave', id, onMapMoveRef.current)
+            map.current.off('mouseleave', id, onMapClickRef.current)
```
justinanderson, line 414: These shouldn't all be `'mouseleave'`, right?

justinanderson (changes requested):
@torosyaneric This definitely fixes the slowdown I was seeing.

There are a few of what I think are typo'd event names. There's no obvious negative effect to them, but it'd be good to get them right.

justinanderson (approved):

PR #298, opened 2022-06-27
Show differences of selected entries(cities for small-business-support) in the comparison page.

leo-esaki:
You can see the comparison page
-through the home page(I alreday set up the 4 entries: 405031120, 406021140, 401011001, 404031105)
-by clicking "View Comparison" button if you added entries in compare menu

### DIFF
--- a/frontend/src/components/ComparisonMenu/ComparisonMenu.js
+++ b/frontend/src/components/ComparisonMenu/ComparisonMenu.js
@@ -14,6 +14,7 @@ import {
 import { useTheme } from '@mui/system'
 import { Delete } from '@mui/icons-material'
 import { useDispatch } from 'react-redux'
+import { useNavigate } from 'react-router'
 import StaticMap from '../StaticMap'
 import { getDatasetGeoJSON } from 'store/modules/dataset'
 
@@ -31,6 +32,7 @@ function ComparisonMenu({
 }) {
   const dispatch = useDispatch()
   const theme = useTheme()
+  const navigate = useNavigate()
   const closeCompareMenuOpen = useCallback(() => setOpen(false), [setOpen])
 
   const handleDeleteClick = useCallback(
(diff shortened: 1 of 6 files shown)


justinanderson (changes requested):
- [ ] Would it help make things faster or simpler if the backend provided an endpoint for getting all compared entries in one request instead of making up to 4 separate requests?

- [ ] Use a back arrow instead of a hamburger menu icon in the comparison page's Header, e.g. Screen Shot 2022-06-27 at 1 49 10 PM

- [ ] When the header is in its alternative scrolled state, each region's thumbnail and name should be aligned over its column. It looks [like this](https://user-images.githubusercontent.com/244952/176005801-acca2deb-e71a-483f-bae4-33109c5dd095.png), but it should look [like this](https://user-images.githubusercontent.com/244952/176005808-383f847b-e449-4434-a86c-b1d2207b29e0.png).

- [ ] Possible to keep the GeoJSON images vertically aligned when the titles have different heights? It currently looks [like this](https://user-images.githubusercontent.com/244952/176005323-55355e68-1cde-48be-9c72-03bfbcb9923a.png), but I'd prefer that look [like this](https://user-images.githubusercontent.com/244952/176005320-195a14c1-b62c-42e1-a155-fe3503e09fd6.png).

- [ ] Clicking a section in the SectionNavbar scrolls the section title below the header, e.g. clicking on "Residential Housing" [scrolls to here](https://user-images.githubusercontent.com/244952/176004453-a7434c56-9746-4076-a8d3-404f3b4b5499.png), but [it should scroll to here](https://user-images.githubusercontent.com/244952/176004455-c428d715-43b9-476e-9d01-689d20340e2b.png).

- [ ] How hard would it be to make controlled sections share a single control, e.g. [not this](https://user-images.githubusercontent.com/244952/176006149-c16b87a1-d950-474b-9784-3e565f4e6780.png) but [this](https://user-images.githubusercontent.com/244952/176006147-eceda15b-d43d-4939-9437-fd62fcd730bf.png).

leo-esaki:
Currently, I figured out 3 problems which I have checked.
- Regarding the first problem, I agree with your suggestion.
- Regarding the third problem, I need to discuss more details with you.
- Regarding the last one, I think it's good to create a separate issue and then implement that.
  It's little tricky and needs complex logic.
My suggestion is to merge the current PR and then figure out remaining things.

